PR #31, opened 2014-07-24
Fixed bugs in embed and remote frame context handling.

RDProjekt:
Hi!

We noticed two bugs in your code. One is in the framing algorithm when the remote frame contains a context url (the "if" was never True and if/else branches were reversed), and the other is in the _remove_embed function (lack of "enumerate" while iterating embed['parent']).

Here's a pull request with patches. We also added some unit-tests for the first bug. Because json-ld-org tests don't cover it, we had to come up with our own.

Best regards,
RD Projekt


mielvds:
This fix and tests are now included in #228

PR #41, opened 2015-05-13
Fixes for running under Python 3

kal:
I found a couple of minor issues when running pyld under Python 3.4.2 which I fixed locally with this PR. I've tried to write the fixes so that they won't break under Python 2.x.


dlongley:
Thanks @kal!

Issue #50, opened 2016-10-14
It is not always relevant to silently ignore unrecognized properties


pchampin:
Consider the following code:
```python
    CTX={ "foo": "http://example.com/foo"}
    expanded = jsonld.expand({"fooo":"bar"}, {'expandContext': CTX})
```
The result is an empty list, because "fooo" is not recognized by CTX. The fact that it is *silently* ignore can be relevant in some situations, but in not in others. In this example, it is probably a typo, which the user would probably like to be detected.

So I think there should be a way to be notified when some JSON attributes were ignored.

The simplest (but not flexible) way to do it would be to have a ``strict`` option, defaulting to False. When set to True, an exception would be raised whenever a JSON key could not be converted to a IRI. In my application, I would rather fail than silently drop some data submitted by the user.

A more flexible solution would be to have a ``droppedKeys`` option, expecting a list, and populating this list with every key that could not be converted to an IRI.
```python
    CTX={ "foo": "http://example.com/foo"}
    dropped = []
    expanded = jsonld.expand({"fooo":"bar"}, {'expandContext': CTX, 'droppedKeys': dropped})
    if dropped:
        # print some warning message to the user
    # but still process expended
```

pchampin:
@dlongley thanks for your comments on fe7eb47.
First, let me just point out what I should have mentionned in the PR: this is just a quick proof of concept. I understand that the actual solution should be thought over more thouroughly.

Regarding "strict mode", I agree that `processingMode`  is a good candidate for playing this role.

Regarding dropped keys, I have indeed overlooked the fact that, with multiple contexts, a key may be converted in some places and dropped in others, so my fix is obviously not satisfactory. Storing the complete path would seem as a valid alternative, but looks harder to achieve in the current code.

I still believe that having a "strict" mode (under this name or another one) would still be a benefit, even if we don't come up with a solution for reporting dropped keys in "normal" mode.

JoelBender:
I believe that a stumbled over this, or maybe `@vocab` can be another way of "not strict".  I expected this document and context...

```
    doc = {
        "@context": {
            "@base": "http://example.org/",
            "@vocab": "http://schema.org/",
            },
        "@id": "12345",
        "@type": "Blog",
        "title": "This is a blog title",
        }
    context = {
        }
```
to generate the same set of triples as this one...

```
    doc = {
        "@id": "12345",
        "@type": "Blog",
        "title": "This is a blog title",
        }
    context = {
[...]

PR #51, opened 2016-10-14
Proposed fix for issue #50

pchampin:


pchampin:
I know that some tests are failing, but they did fail (on my machine at least) on branch master as well, so I do not think that this is due to my patch...

lib/pyld/jsonld.py
```diff
@@ -1979,6 +1982,10 @@ def _expand(
                 not (
                     _is_absolute_iri(expanded_property) or
                     _is_keyword(expanded_property))):
+                if options['strict']:
+                    raise ValueError("Unrecognized key %s" % key)
+                elif droppedKeys is not None:
+                    droppedKeys.add(key)
```
anatoly-scherbakov, line 1988: Returning a value from a function by modifying an argument is quite an unpopular method with Python programmers. This is a side effect; I would discourage it.

anatoly-scherbakov (changes requested):

lib/pyld/jsonld.py
BigBlueHat: @anatoly-scherbakov could you refactor this in a way that still fixes the issue?

pchampin: @anatoly-scherbakov I disagree that it is inherently a bad thing for a function to mutate one of its arguments, as long as the contract is clear about it. 
That being said, I agree that having this "output argument" included in the "options" argument (which is clearly an input argument otherwise), is not great.

I don't consider that changing the return value of `expand`, in order to include the information about dropped keys, is desirable (it would break existing code).

I believe that a better solution to be to add a 4th parameter to expand: `dropped=None`. If that parameter is explicitly specified, it must be an (empty?) list, and the contract would be that this list would be populated with the dropped keys.

anatoly-scherbakov: Even if the contract is explicitly specified, this is not what a Python developer is _expecting_ from a library, thus violating the [Principle of Least Astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

I have proposed a PR: https://github.com/digitalbazaar/pyld/pull/186 — but its not ready for merge as it should be really targeted at another PR of mine.

PR #55, opened 2017-03-30
Check if parent is @list, closes #54

### LINKED ISSUE #54: Framing of @container:@list with embedded reference errors
The code below errors:
```
  ...
  File "/Library/Python/2.7/site-packages/pyld/jsonld.py", line 3320, in remove_dependents
    embeds[next]['parent']['@id'] == id_):
KeyError: '@id'
```

When the `ref` keys are present. 
Expected behavior would be to produce the same result as the input document.
Version: 0.7.1

```python

from pyld.jsonld import frame

ctxtjs = {
	"@context": {
		"ns": "http://example.com/ns/",
		"hasList": {"@id": "ns:hasList", "@type": "@id", "@container":"@list"},
		"ref": {"@id": "ns:ref", "@type": "@id"}
	}	
}

doc = {
	"@context": ctxtjs['@context'],
	"@id": "http://top/1",
	"@type": "ns:A",
	"hasList": [
		{
			"@id": "http://child/1", 
		 	"@type": "ns:B", 
		 	"ref": "ns:something"
		},
		{
			"@id": "http://child/2", 
		 	"@type": "ns:B", 
		 	"ref": "ns:something"
		}
	]
}

framejs = {
	"@context": ctxtjs['@context'],
	"@type": "ns:A",
	"hasList": [{
		"@type": "ns:B"
	}]
}

framed = frame(doc, framejs)
```

(/cc @zimeon, @jpstroop, re IIIF/iiif.io#1117)

azaroth42:
If the parent resource is an `@list`, it does not have `@id`, therefore line 3321 fails with the `KeyError`. This patch adds in a check that `@id` is indeed in the parent keys.

Closes #54, but it would be good to add the code from the issue as a test case.

### DIFF
--- a/lib/pyld/jsonld.py
+++ b/lib/pyld/jsonld.py
@@ -3316,8 +3316,6 @@ class JsonLdProcessor(object):
             except AttributeError:
                 ids = list(embeds.keys())
             for next in ids:
-                print next
-
                 if (next in embeds and
                         _is_object(embeds[next]['parent']) and
                         '@id' in embeds[next]['parent'] and # could be @list
--- a/lib/pyld/jsonld.py
+++ b/lib/pyld/jsonld.py
@@ -3316,8 +3316,11 @@ class JsonLdProcessor(object):
             except AttributeError:
                 ids = list(embeds.keys())
             for next in ids:
+                print next
+
                 if (next in embeds and
                         _is_object(embeds[next]['parent']) and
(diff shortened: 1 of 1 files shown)


azaroth42:
Ping? We're blocked by this bug and would really prefer not to fork a new module to pypi.

davidlehn:
Released in 0.7.2.

PR #60, opened 2017-07-18
Always try to add a default @type and @language

### LINKED ISSUE #59: Dates from schema.org are not compacted correctly.
I've traced this down to https://github.com/digitalbazaar/pyld/blob/master/lib/pyld/jsonld.py#L4132

The original symptoms that I observed were that various `Date` fields that should be compacted to keys like `birthDate` were actually including a `schema:` prefix.

It looks like the core problem is that the default inverse dictionary for `@language` is not getting populated correctly (perhaps due to a change in the published `@context` from schema.org?)

Date field specifications include both `@id` and `@type`, but not language, and are still expected to be pure strings ([according to schema.org's documentation](http://schema.org/Date)).  When the compactor attempts to find the correct term, `[_select_term]()` returns None. 

The inverse dictionary winds up looking something like this:

```python
{
    'http://schema.org/birthDate': {'@none': {'@language': {},
                                              '@type': {'http://schema.org/Date': 'birthDate'}}},
    # etc
}
```

Instead of containing a `{'@none': 'birthDate'}` for `@language`, there's just an empty dict.

Some other non-date fields also seem to exhibit this issue, but I don't know enough about the library or json-ld to know if these symptoms are actually problems, or if they're by design.

Minimum-reproducible sample:

```python
#!/usr/bin/env python

from pyld import jsonld

doc = {
    'http://schema.org/name': 'Buster the Cat',
    'http://schema.org/birthDate': '2012',
    'http://schema.org/deathDate': '2015-02-25'
}

frame = {
    '@context': 'http://schema.org/'
}

framed = jsonld.frame(doc, frame)
contents = framed['@graph'][0]
print(framed)
assert 'name' in contents  # fine
assert 'birthDate' in contents  # not fine, schema:birthDate instead
assert 'deathDate' in contents  # not fine, schema:deathDate instead
```

My proposal to fix this would be to apply https://github.com/Artory/pyld/commit/faaa1394dc32ecfdd6fd875f45e4fdbc931fa7cc, to attempt to set these defaults regardless of the outcome of the conditionals there.

rknLA:
Fixes #59

In cases where @type was set, but not @language, one or the other would end up being an empty dictionary.

I have not run any tests on this, or added test cases, so that might need to be addressed before merging.

### DIFF
--- a/lib/pyld/jsonld.py
+++ b/lib/pyld/jsonld.py
@@ -4182,9 +4182,9 @@ class JsonLdProcessor(object):
                 else:
                     # add an entry for the default language
                     entry['@language'].setdefault(default_language, term)
-                    # add entries for no type and no language
-                    entry['@type'].setdefault('@none', term)
-                    entry['@language'].setdefault('@none', term)
+                # add entries for no type and no language
+                entry['@type'].setdefault('@none', term)
+                entry['@language'].setdefault('@none', term)
 
         return inverse
 


lib/pyld/jsonld.py
```diff
@@ -4182,9 +4182,9 @@ def _get_inverse_context(self, active_ctx):
                 else:
                     # add an entry for the default language
                     entry['@language'].setdefault(default_language, term)
-                    # add entries for no type and no language
-                    entry['@type'].setdefault('@none', term)
-                    entry['@language'].setdefault('@none', term)
+                # add entries for no type and no language
+                entry['@type'].setdefault('@none', term)
+                entry['@language'].setdefault('@none', term)
```
anatoly-scherbakov, line 4187: I don't know but I am concerned about verbosity of the resulting document. What do you think about this?

PR #63, opened 2017-09-12
fix variables in jsonld.link

bpow:
The variable `frame` was masking scope of jsonld.frame. Also, `input` was being passed to jsonld.frame instead of the parameter `input_`.


davidlehn:
Thanks.  Sigh... I'm guessing no one had ever run that code before.  It was likely just ported from js without proper testing.

0.7.3 released on PyPI.

Issue #65, opened 2017-09-13
HTTPS SNI Verification fails in python 2


jkreft-usgs:
this is related to #57 
Running on python 2.7.11 and attempting to load a remote context url hosted on github.io.  It seems that the HTTPS handler that was included in this package is circumventing the native HTTPS handling backported into python past 2.7.9.  

This is the stacktrace:
```JsonLdError: ('Could not process context before compaction.',)
Type: jsonld.CompactError
Cause: ('Could not process JSON-LD context.',)
Type: jsonld.ContextError
Cause: ('Dereferencing a URL did not result in a valid JSON-LD context.',)
Type: jsonld.ContextUrlError
Code: loading remote context failed
Details: {'url': <github.io url>}
Cause: ('Could not retrieve a JSON-LD document from the URL.',)
Type: jsonld.LoadDocumentError
Code: loading document failed
Cause: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)>  File "/Users/user/venvs/env/lib/python2.7/site-packages/pyld/jsonld.py", line 390, in load_document
    with closing(url_opener.open(url)) as handle:
  File "/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 431, in open
    response = self._open(req, data)
  File "/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 449, in _open
    '_open', req)
  File "/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 409, in _call_chain
    result = func(*args)
  File "/Users/user/venvs/env/lib/python2.7/site-packages/pyld/jsonld.py", line 5135, in https_open
    return self.do_open(self.specialized_conn_class, req)
  File "/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/urllib2.py", line 1197, in do_open
    raise URLError(err)
  File "/Users/user/venvs/env/lib/python2.7/site-packages/pyld/jsonld.py", line 4072, in _retrieve_context_urls
    remote_doc = load_document(url)
  File "/Users/user/venvs/env/lib/python2.7/site-packages/pyld/jsonld.py", line 430, in load_document
    cause=cause)
[...]

Issue #57, opened 2017-06-20
SSL certificate verify failures in tests


davidlehn:
The tests that access https://json-ld.org/ have "SSL: CERTIFICATE_VERIFY_FAILED" failures.  Perhaps due to the local VerifiedHTTPSHandler not supporting SNI?

mielvds:
As far as I can tell, this was resolved by introducing requests in #67 about 9 years ago, so closing!

PR #67, opened 2017-10-11
requests, paths, etc

davidlehn:
Looking for feedback on the doc loader API:
-  Should it more closely match the jsonld.js one?  (ie, a documentLoader.requests object or similar)
-  This one doesn't have to support web and servers but it's still a similar pattern that allows using custom loaders.  There might be a use case for changing things to support the old API vs requests lib.
- Is it worth the trouble to made 'requests' optional and provide a simpler (though maybe broken) loader like the old code?

The path futzing changes pass the tests now but it's unclear how correct or optimal it is.  I just moved closer to what jsonld.js does.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,6 @@ python:
   - "3.6"
   - "pypy3"
 sudo: false
-# no requirements, but simulate what a user would see
-install: pip install .
 # download test suite and run tests... submodule? meta testing project with
 # all of the reference implementations?
 script: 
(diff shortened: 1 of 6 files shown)


davidlehn:
Addresses:
- https://github.com/digitalbazaar/pyld/issues/57
- https://github.com/digitalbazaar/pyld/issues/65

Elizafox:
requests is quite a common dependency so I see no harm in depending on it.

jkreft-usgs:
It would be pretty sweet if this PR could be merged and released soon.  I have several projects that could take advantage of pyld it right away.

davidlehn:
Ok, I'll merge and release. Please file issues for bugs or suggestions. We have been using jsonld.js much more than this lib lately so do appreciate feedback for things we may miss beyond what's in the test suite.

PR #73, opened 2018-01-26
Support for Asynchronous Document Loader using aiohttp

oohlaf:


davidlehn:
Cool!  I thought it would be more invasive to do such things.  The README.rst also needs to be updated (requirements, installation notes, doc loader section, etc).  Also would by nice to update travis config so tests could pass on older pythons using requests.  I've gotten used to semantic versioning from js world.  Probably need to bump the minor when this is released.

Also sorry the tests have many failures.  The code needs to be updated for new tests that were added.

oohlaf:
I thought so as well, but then I found out I could wrap the loader in a run_until_complete task. I can update the readme and tox probably as well. Need to look into how to define conditions in tox. Will give it a go somewhere next week.

oohlaf:
Also, this version will only allow other async tasks to run whenever remote URLs need to be fetched, otherwise it will execute the entire compact or expand command in one go.

oohlaf:
David, let me know if you're ok with splitting the document loaders to separate files before I update the main README file. I had to split it off, otherwise older Python versions keep throwing SyntaxErrors in the main file.

davidlehn:
Yes, splitting out document loaders is fine.  There is some work going on in the `json-ld-1.1` branch.  It might involve splitting up all the code similar to what was done in jsonld.js.  The document loaders were extracted there too.

oohlaf:
Hi, I've updated the documentation and moved the document loaders to a separate folder.
I've also rebased the commits to 3 logical ones (creation of aiohttp, travis changes, and moving to separate folder). Noticed that there was a name clash on Python2.7 because that version does not do absolute imports. Luckily that can be turned on as __future__.

In #77 I did a PR against the json-ld-1.1 branch as that branch contains a functioning test suite.
This PR #73 is against the master branch where the test suite still fails (even though it says passed above, but the GitHub Travis component references the wrong build number. On the Travis website you can see it failed).

You're free to decide in which branch you'd like to merge :wink:

davidlehn:
Thanks for doing this!  We're a bit busy at the moment and will try to merge and release the 1.1 code and aiohttp patch soon.

Any help reviewing or testing the 1.1 PRs would be great too.  We don't use the py code as much as the js version so may miss some issues that the test suite doesn't catch.

davidlehn:
Thanks!  This will be in next release along with 1.1 draft spec features.

PR #75, opened 2018-02-14
Support for 1.1 framing.

gkellogg:
This brings pyld support for JSON-LD 1.1 up to par with jsonld.js. Given that the json-ld-1.1 branch is (AFAIK) unreviewed, this arguably doesn't make the reviewing situation any worse, so it could just be merged into the json-ld-1.1 branch, but it would be good to have more critical eyes on both of these branches at some point.

### DIFF
--- a/lib/pyld/jsonld.py
+++ b/lib/pyld/jsonld.py
@@ -3888,9 +3888,9 @@ class JsonLdProcessor(object):
             return True
         if not (v1 in v2 or _is_empty_object(v2[0])):
             return False
-        if not (not t1 and not t2 or t1 in t2 or t1 and t2 and _is_empty_object(t2[0])):
+        if not ((not t1 and not t2) or (t1 in t2) or (t1 and t2 and _is_empty_object(t2[0]))):
             return False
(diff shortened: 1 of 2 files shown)


gkellogg:
Some Python 2.7 and 3.4 issues to resolve.

lib/pyld/jsonld.py
```diff
@@ -3644,6 +3850,51 @@ def _add_frame_output(self, parent, property, output):
...
+        v1, t1, l1 = value.get('@value'), value.get('@type'), value.get('@language')
+        v2 = JsonLdProcessor.get_values(pattern, '@value')
+        t2 = JsonLdProcessor.get_values(pattern, '@type')
+        l2 = JsonLdProcessor.get_values(pattern, '@language')
+
+        if not v2 and not t2 and not l2:
+            return True
+        if not (v1 in v2 or _is_empty_object(v2[0])):
+            return False
+        if not (not t1 and not t2 or t1 in t2 or t1 and t2 and _is_empty_object(t2[0])):
```
davidlehn: Maybe add some parens for readability?

PR #76, opened 2018-02-14
Support JSON-LD 1.1

davidlehn:


azaroth42:
@dlongley @davidlehn Any timeline for when we might be able to merge to master?

dlongley:
@davidlehn,

If we're passing all the tests -- and we have the same fix we put in for compact `@graph` arrays as in jsonld.js then I vote we go ahead and merge and do a `1.0.0` release of pyld.

davidlehn:
I'm ok for release as well once that fix is in.  The js patch in question is this:
https://github.com/digitalbazaar/jsonld.js/commit/4ce2b708a012ef1fa3e6f5f6b7e35950e6a98b38

There was also a question if the above fix works with `compactArrays` api option.  I haven't had time to check if we have tests for that and/or write new ones.

gkellogg:
I’m out of the country and can’t respond until next week.

azaroth42:
Thank you! :)

Issue #78, opened 2018-03-08
regression? JsonLdError: <unprintable JsonLdError object>


yarikoptic:
Our @datalad testing build bots started to puke a new error which we haven't seen before:
```
======================================================================
ERROR: datalad.metadata.tests.test_search.test_our_metadataset_search
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 133, in run
    self.runTest(result)
  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 151, in runTest
    test(result)
  File "/usr/lib/python2.7/unittest/case.py", line 393, in __call__
    return self.run(*args, **kwds)
  File "/usr/lib/python2.7/unittest/case.py", line 353, in run
    result.addError(self, sys.exc_info())
  File "/usr/lib/python2.7/dist-packages/nose/proxy.py", line 128, in addError
    formatted = plugins.formatError(self.test, err)
  File "/usr/lib/python2.7/dist-packages/nose/plugins/manager.py", line 99, in __call__
    return self.call(*arg, **kw)
  File "/usr/lib/python2.7/dist-packages/nose/plugins/manager.py", line 141, in chain
    result = meth(*arg, **kw)
  File "/usr/lib/python2.7/dist-packages/nose/plugins/logcapture.py", line 237, in formatError
    return (ec, self.addCaptureToErr(ev, records), tb)
  File "/usr/lib/python2.7/dist-packages/nose/plugins/logcapture.py", line 244, in addCaptureToErr
    records + \
  File "/usr/lib/python2.7/dist-packages/nose/util.py", line 652, in safe_str
    return str(val)
  File "/home/buildbot/datalad-pr-docker-dl-nd90/build/venv-ci/local/lib/python2.7/site-packages/pyld/jsonld.py", line 4923, in __str__
    rval += '\nCause: ' + str(self.cause)
  File "/home/buildbot/datalad-pr-docker-dl-nd90/build/venv-ci/local/lib/python2.7/site-packages/pyld/jsonld.py", line 4923, in __str__
    rval += '\nCause: ' + str(self.cause)
  File "/home/buildbot/datalad-pr-docker-dl-nd90/build/venv-ci/local/lib/python2.7/site-packages/pyld/jsonld.py", line 4923, in __str__
    rval += '\nCause: ' + str(self.cause)
[...]

PR #79, opened 2018-03-08
Fix: Provide jsonld.LoadDocumentError for the type of error in dummy_document_loader

yarikoptic:
Addresses a part of #78 forbidding proper error msg to be displayed


davidlehn:
Thanks!

PR #81, opened 2018-04-27
Framing - reset graph context after recursing through named graphs

pjohnston-wiley:
When framing a JSON-LD document with at least one named graph, the current graph (`state['graph']`) was not being reset after recursing through the named graphs. Since the code iterates in alphabetical order, this only manifested for objects in the default graph that had `@id`s greater than the `@id` of the first named graph. This would be a good test to put into the overall JSON-LD framing test suite, please let me know if you want a formal example.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ cover
 dist
 docs/_build
 lib/PyLD.egg-info
+.idea
--- a/lib/pyld/jsonld.py
+++ b/lib/pyld/jsonld.py
@@ -3458,6 +3458,8 @@ class JsonLdProcessor(object):
                     # recurse into graph
                     self._match_frame(
                         state, sorted(state['graphMap'][id_].keys()), [subframe], output, '@graph')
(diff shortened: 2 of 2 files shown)


davidlehn:
Hi, If you have a chance could you provide a test example?  I'll try to get it into main json-ld.org test suite to avoid the same issue in other implementations.  If you want to add a patch to that repo, that works too.  Thanks, and apologies for the late response.

pjohnston-wiley:
sure - i may already have one in some of the posts i put on the WG issues list, e.g. [this](https://github.com/json-ld/json-ld.org/issues/641). I'm traveling this week but i'll try and put something together soon.

pjohnston-wiley:
[There now](https://github.com/json-ld/json-ld.org/pull/663)

anatoly-scherbakov:
Should this be merged since https://github.com/json-ld/json-ld.org/pull/663 is merged?

Issue #86, opened 2018-08-22
The JsonLdProcessor.parse_nquads method fails to parse an nquad line with an uppercase region subtag


apeters:
For example this nquad line will fail to parse
```
'<http://www.test.org> <http://purl.org/dc/terms/identifier> "http://www.test.org/id1"@en-US <http://test.org/test> .'
```

notice the use of `@en-US` 
the line will parse just fine if `@en-us` is used

generally, it seems that the region subtag (the `US` part) is usually displayed as uppercase.
see: https://www.w3.org/International/articles/language-tags/#region

But more importantly, the case of the entire language tag should be ignored (at least as far as parsing is concerned).

Per: https://tools.ietf.org/html/rfc5646#section-2.1.1
```
 At all times, language tags and their subtags, including private use
 and extensions, are to be treated as case insensitive: there exist
 conventions for the capitalization of some of the subtags, but these
 MUST NOT be taken to carry meaning.
```

davidlehn:
1.0.4 released with #87.

PR #87, opened 2018-08-22
when parsing nquads, allow the language tag to be of mixed case

apeters:
re #86


azaroth42:
@dlongley @davidlehn Any chance that this quick one could be merged? It's blocking us for things like `en-US` (which everyone uses) versus `en-us` (which is accepted by the code but no one does in practice).  Thanks!

davidlehn:
Ok.  We have that in js code already: https://github.com/digitalbazaar/rdf-canonize/commit/095c1747304186e1ba8267daeee0d2751b22a2f8
I'm ok with this, I just wonder if there is confusion with some normalization needing to happen if people mix upper and lower case. Will deal with that when it's a real issue.

davidlehn:
1.0.4 released.

PR #93, opened 2019-05-09
Fix behavior to terminate generators

jevans97utk:
Many more tests fail when running the normalization test suite on Python
3.7 than on 2.7, 3.5, or 3.6.  The reason is PEP 479, which introduces a
change in behavior when raising StopIteration inside a generator.  The
new behavior is to silently turn this into a RuntimeError.

The PEP also states the way to deal with this, i.e.

    Finally, the proposal also clears up the confusion about how
    to terminate a generator: the proper way is return, not raise
    StopIteration.

### DIFF
--- a/lib/pyld/jsonld.py
+++ b/lib/pyld/jsonld.py
@@ -5499,7 +5499,7 @@ def permutations(elements):
 
         # no more permutations
         if k is None:
-            raise StopIteration
+            return
 
         # swap k and the element it is looking at
         swap = pos - 1 if left[k] else pos + 1


davidlehn:
Can you rebase this on master?  I just updated travis config to test on 3.7.  Want to make sure tests work the same in all versions with this patch.  Pardon all the other failing tests, pyld has fallen behind the js code as far as keeping the test runner itself up-to-date, and both are behind in the latest 1.1 spec updates.

davidlehn:
Thanks. 1.0.5 released with the fix.

PR #106, opened 2020-03-08
1.1 API

gkellogg:
* **BREAKING**: Versions of Python before 3.4 are no longer supported.
* Update conformance docs.
* Add all keywords and update options.
* Default processingMode to json-ld-1.1.
* Implement logic for marking tests as pending, so that it will fail if a pending test passes.
* Consolidate documentLoader option and defaults into a load_document method to also handle JSON (eventually HTML) parsing.
* Add support for rel=alternate for non-JSON-LD docs.
* Context processing:
  * Support `@propagate` in context processing and propagate option.
  * Support for `@import`. (Some issues confusing recursion errors for invalid contexts).
  * Make `override_protected` and `propagate` optional arguments to `_create_term_definition` and `_process_context` instead of using option argument.
  * Improve management of previous contexts.
  * Imported contexts must resolve to an object.
  * Do remote context processing from within _process_contexts, as logic is too complicated for pre-loading. Removes _find_context_urls and _retrieve_context_urls.
* Create Term Definition:
  * Allow `@type` as a term under certain circumstances.
  * Reject and warn on keyword-like terms.
  * Support protected term definitions.
  * Look for keyword patterns and warn/return.
  * Look for terms that are compact IRIs that don't expand to the same thing.
  * Basic support for `@json` and `@none` as values of `@type`.
  * If `@container` includes `@type`, `@type` must be `@id` or `@vocab`.
  * Support `@index` and `@direction`.
  * Corner-case checking for `@prefix`.
  * Validate scoped contexts even if not used.
  * Support relative vocabulary IRIs.
  * Fix check that term has the form of an IRI.
  * Delay adding mapping to end of _create_term_definition.
  * If a scoped context is null, wrap it in an array so it doesn't seem to be undefined.
* IRI Expansion:
  * Find keyword patterns.
  * Don't treat terms starting with a colon as IRIs.
  * Only return a resulting IRI if it is absolute.
  * Fix _is_absolute_iri to use a reasonable regular expression and some other _expand_iri issues.
  * Fix to detecting relative IRIs.
  * Fix special case where relative path should not have a leading '/'
* Compaction:
  * Compact `@direction`.
  * Compact `@type`: `@none`.
  * Compact `@included`.
  * Honor `@container`: `@set` on `@type`.
  * Lists of Lists.
  * Improve handling of scoped contexts and propagate.
  * Improve map compaction, including indexed properties.
  * Catch Absolute IRI confused with prefix.
* Expansion:
  * Updates to expansion algorithm.
  * _expand_value adds `@direction` from term definition.
  * JSON Literals.
  * support `@direction` when expanding.
  * support lists of lists.
  * support property indexes.
  * improve graph container expansion.
  * Order types when applying scoped contexts.
  * Use type_scoped_ctx when expanding values of `@type`.
  * Use propagate and override_protected properly when creating expansion contexts.
* Flattening:
  * Rewrite _create_node_map based on 1.1 algorithm.
  * Flatten `@included`.
  * Flatten lists of lists.
  * Update merge_node_maps for `@type`.
* ToRdf:
  * Support for I18N direction.
  * Support for Lists of Lists.
  * Partial support for JSON canonicalization of JSON literals.
    * Includes local copy of JCS library, but doesn't load.
  * Lists of Lists.
  * Text Direction 'i18n-datatype'.


davidlehn:
There's also a CHANGELOG.md file. I'm not sure what level of detail should eventually go in there for such a large batch of changes.  At least something like "Support JSON-LD 1.1.". ;-)  Probably also add any important API changes or similar things that 1.0 users would need to know when upgrading.

gkellogg:
I figured that I would use the PR summary for the changelig when I’m closer to being ready for a merge. 

It’s going pretty well, and now need to whittle away at the remaining pending expansion tests, as I’ve made a pass through all the relevant algorithms.

gkellogg:
This does everything for expansion other than HTML and some remote-doc tests which have probably never worked properly.

Some things I can't resolve and will require a change in approach:

* tests so01, so03, and so04 should get an "invalid context entry" but instead get "recursive context inclusion". This is because of pre-fetching. It seems to me that jsonld.js must have the same issue, but seems to pass these tests.
* test js21 references an unused context, but it fails when loading, similar issue.

README.rst
```diff
@@ -42,39 +42,15 @@ Conformance
...
 
-- `JSON-LD 1.0`_,
-  W3C Recommendation,
-  2014-01-16, and any `errata`_
-- `JSON-LD 1.0 Processing Algorithms and API <JSON-LD 1.0 API_>`_,
-  W3C Recommendation,
-  2014-01-16, and any `errata`_
-- `JSON-LD 1.0 Framing <JSON-LD 1.0 Framing_>`_,
-  Unofficial Draft,
-  2012-08-30
```
davidlehn: The 1.1 support update should still work for most 1.0 data right?  There are some exceptions but I didn't think enough to remove this text.  I imagine the old framing and 1.1 CG draft support is more questionable at this point.

lib/pyld/jsonld.py
```diff
@@ -4697,9 +5242,21 @@ def _find_context_urls(self, input_, urls, replace, base):
                     elif v not in urls:
                         urls[v] = False
                 elif _is_object(v):
-                    # look for soped context
+                    # look for sooped context
```
davidlehn: sooped context... nomnom.
```suggestion
                    # look for scoped context
```

davidlehn:
I do wish we had split that file up similar to jsonld.js.  Kind of a monster to look at now!

README.rst
gkellogg: The 1.1 specs are compatible with the 1.0 steps; it's redundant to list all previous versions of specifications supported. This includes Framing 1.1 being compatible with the unofficial 1.0 draft.

gkellogg:
@davidlehn I hit a wall with both remote and local installation of the JCS module. I've included it here, but it's not loading, and I really don't understand how to manage this environment. Any help you can provide would be most useful.

I commented out the module include and the use within `_object_to_rdf`, just using `json.dumps` for now.

davidlehn:
I got it to try and do remote install, but the lack of a setup.py seems to be a dealbreaker.
```sh
pip install "git+https://github.com/cyberphone/json-canonicalization@571690e6c9af5f614a29bc0e6034fafec93f34c8#egg=org.webpki.json.Canonicalize&subdirectory=python3"
```
Not sure if that's correct (I'm guessing), but fails due to lack of setup.py.  Would be a `[-e]` line in requirements.txt if it could be installed.  I think we'll have to deal with a copy for now.

.travis.yml
```diff
@@ -2,8 +2,6 @@ dist: xenial
 language: python
 cache: pip
 python:
-  - "2.7"
-  - "pypy"
   - "3.4"
```
gkellogg: Probably want to just require 3.5 or better, as I've introduced some code and programming patterns that require it. (other stuff should be cleaned up).

README.rst
```diff
@@ -83,7 +59,7 @@ yet supported.
 Requirements
 ------------
 
-- Python_ (2.7 or later)
+- Python_ (3.4 or later)
```
gkellogg: 3.5 or better, I think.

tests/runtests.py
```diff
@@ -241,6 +233,12 @@ def setUp(self):
             if re.match(regex, data.get('@id', data.get('id', ''))):
                 self.skipTest('Test with id regex %s' % regex)
 
+        # mark tests as pending, meaning that they are expected to fail
```
gkellogg: Better than skipping tests that aren't ready, as it will fail when it starts passing.

tests/runtests.py
```diff
@@ -485,6 +567,13 @@ def __init__(self):
             'doap:license': 'https://github.com/digitalbazaar/pyld/blob/master/LICENSE',
             'doap:description': 'A JSON-LD processor for Python',
             'doap:programming-language': 'Python',
+            'doap:release': {
+                'doap:revision': '2.0.0', # FIXME: this should come from a file
```
gkellogg: This should come out of a file, obviously.

gkellogg:
A comprehensive EARL report requires running both /api and /framing test suites and merging the result using some external tooling. Could just submit as N-Triples or Turtle by converting JSON-LD output.

gkellogg:
@davidlehn thanks for updating the test runner, much easier now.

PR #107, opened 2020-03-19
Fix canonicalize loading.

davidlehn:
- Remove unrelated 'canonicalize' package from requirements.txt.
- Fix package namespaces.
- Incorrect fix for value type.

That str(..) thing is wrong, I forget the right way to do that, but this is a start.

### DIFF
--- a/lib/c14n/Canonicalize.py
+++ b/lib/c14n/Canonicalize.py
@@ -22,7 +22,7 @@
 
 import re
 
-import convert2Es6Format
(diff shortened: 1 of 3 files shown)


davidlehn:
Partial fix for https://github.com/digitalbazaar/pyld/pull/106.

gkellogg:
Thanks @davidlehn, I'll see if that helps.

gkellogg:
`canonicalize(value).decode('UTF-8')` does it.

Thanks so much.

PR #108, opened 2020-03-22
Load from HTML

gkellogg:
* Use lxml.html to load HTML and parse in load_html.
  * For HTML, the API base option can be updated from base element.

Note this is a PRI into 1.1-expansion.


gkellogg:
Note the 3.4 failure:

> Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-install-phouehag/lxml/

Do we want to bump the minimum version to 3.5 or figure out how to do this conditionally?

dlongley:
@davidlehn, please review.

PR #112, opened 2020-03-30
ContextResolver.

gkellogg:


gkellogg:
There are about four errors to resolve, having to do with 'protected' propagation and properly remembering the base for the processing context.

gkellogg:
@azaroth42 @kasei, this might be worth checking out to see how it affects your processing times, in spite of the remaining errors.

azaroth42:
We don't use protected branches, so if that's the only challenge, then it should be a great benefit! 

@kasei, can we try this out with a run tonight or tomorrow and see what happens?

kasei:
@gkellogg does the `c14n` module directory need to be listed in the setup.py `packages` list? I'm not seeing it get installed along with the rest of the package using pip.

gkellogg:
@davidlehn got it this far, it seems to work enough for Travis to find it. I'm really not a Python person (in spite of all the work to the contrary) and the ecosystem is something of a mystery to me. Perhaps there's something to be gleaned from the .travis.yml.

kasei:
I also don't feel much expertise in this area. However, I suspect that travis doesn't run into this problem, because it isn't installing pyld. Instead, it's just installing the prerequisites, and then testing the local copy of pyld.

I'll also mention that there seems to be some issue with cachetools being listed as a prerequisite. It didn't install for me when I installed using pip via a git URL. I'm not sure what the issue is, but I had to install it manually.

gkellogg:
@davidlehn Probably something needs to go in setup.py? A package description?

```python
setup(
    name='PyLD',
    version=about['__version__'],
    description='Python implementation of the JSON-LD API',
    long_description=long_description,
    author='Digital Bazaar',
    author_email='support@digitalbazaar.com',
    url='http://github.com/digitalbazaar/pyld',
    packages=[
        'pyld',
        'pyld.documentloader',
        'c14n'
    ],
    package_dir={'': 'lib'},
    license='BSD 3-Clause license',
    classifiers=[
        'Development Status :: 4 - Beta',
        'Environment :: Console',
        'Environment :: Web Environment',
        'Intended Audience :: Developers',
        'License :: OSI Approved :: BSD License',
        'Operating System :: OS Independent',
        'Programming Language :: Python',
        'Topic :: Internet',
        'Topic :: Software Development :: Libraries',
    ],
    install_requires=[],
    extras_require= {
        'requests': ['requests'],
        'aiohttp': ['aiohttp'],
        'cachetools': ['cachetools']
    }
)
```

'c14n' in packages and 'cachetools' in extras_require?

azaroth42:
Yep, when I add `'c14n'` to packages, then it will install for me. I already have cachetools installed, so not sure whether that's needed or not. It should be in `requirements.txt` though, rather than the package file, I think... but also not super familiar with the ins and outs of pypi package management!

Then in order to reuse a context resolver, I need to add ContextResolver to `__all__` in jsonld.py

azaroth42:
Then I can do this:
```
from pyld.jsonld import compact, expand, set_document_loader, ContextResolver
import json
from cachetools import LRUCache

resolved_context_cache = LRUCache(maxsize=1000)
docCache = {}

fh = open('linked-art.json')
js = json.load(fh)
fh.close()
docCache["https://linked.art/ns/v1/linked-art.json"] = \
	{
		"contextUrl": None,
		"documentUrl": "https://linked.art/ns/v1/linked-art.json",
		"document": js
	}

def fetch(url):
	resp = requests.get(url)
	return resp.json()

def load_document_and_cache(url, options={}):
	if url in docCache:
		return docCache[url]
	doc = {"contextUrl": None, "documentUrl": url, "document": ""}
	doc["document"] = fetch(url)
	docCache[url] = doc
	return doc

set_document_loader(load_document_and_cache)
resolver = ContextResolver(resolved_context_cache, load_document_and_cache)

indata = {"@context":"https://linked.art/ns/v1/linked-art.json", "id": "tag:foo", "type": "Person"}
e = expand(indata, options={'contextResolver': resolver})
```

And at the end of execution, the cache has a bunch of bits of context in it. Haven't tested to see if they get reused correctly, or if there's any improvement to speed of processing.

gkellogg:
cachetools is in requirements.txt, but I'll add other things that seem to be necessary; @dlongley or @davidlehn can always comment in their review before merge.

> Yep, when I add `'c14n'` to packages, then it will install for me. I already have cachetools installed, so not sure whether that's needed or not. It should be in `requirements.txt` though, rather than the package file, I think... but also not super familiar with the ins and outs of pypi package management!
> 
> Then in order to reuse a context resolver, I need to add ContextResolver to `__all__` in jsonld.py

Seems like it should be `__all__` in `__init__.py`, not `jsonld.py`, as jsonld.py doesn't expose it.

I'll push a commit that I think addresses both ContextResolver and Canonicalization.

azaroth42:
It can expose it, because it imports it from context_resolver.  From a module API perspective, I think having it exposed from jsonld is cleaner, as it has *all* of the other functionality, spread across multiple classes and functions. Remembering which classes are in which files in this case would be a pain.

gkellogg:
For some reason, the travis results don't show here. Three tests fail in Python 3.5, but pass in 3.6, 3.7 and pypy. Needs more investigation.

azaroth42:
I think it should be:

```
+from .Canonicalize import canonicalize

+from .context_resolver import ContextResolver
+from .resolved_context import ResolvedContext
```

to get it to reference the classes in the right files

gkellogg:
> I think it should be:
> 
> ```
> +from .Canonicalize import canonicalize
> 
> +from .context_resolver import ContextResolver
> +from .resolved_context import ResolvedContext
> ```
> 
> to get it to reference the classes in the right files


`canonicalize` is in `c14n.Canonicalize`, trying to use `.Canonicalze` fails "No module named 'pyld.Canonicalize'".

The other two are in jsonld.py.

azaroth42:
sorry, that first should be in c14n.__init__

gkellogg:
@azaroth42 as in https://github.com/digitalbazaar/pyld/pull/112/commits/2711f57bbab97adbcb3b7061a3e01fea5c6b1762 and https://github.com/digitalbazaar/pyld/pull/112/commits/932bbe0ba5e1461e4bcb26a807c5b163938d4ea8?

azaroth42:
Some minimal wall-clock benchmarking makes me think there's a fundamental issue somewhere in the code.  

I have 150 different instances (enumerated here https://linked.art/example/index.json) and using the ContextResolver it takes 30 seconds to expand all of them.  Using master, it takes just 3 seconds! 

I haven't done profiling yet to see where it is spending so much time yet.

gkellogg:
It would certainly be worth checking to see if the ContextResolver is preventing access to those files. Also, how did you determine that time is spent parsing contexts? Perhaps it's somewhere else in the path.

It does seem that a profile is necessary.

azaroth42:
The timing wraps only the call to `expand()`.  

```
start = time.time()
for indata in all_data:
	e = expand(indata)
print(time.time() - start)
```

I haven't tried with your other 1.1 branch, I admit. So if this branch builds from that one, then it could be in some other path, I agree. Will report back (tomorrow) with profiling data.

azaroth42:
The vast, vast majority of the time is spent deepcopying things:

```
      150    0.001    0.000   63.200    0.421 /Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pyld/jsonld.py:137(expand)
      150    0.001    0.000   63.198    0.421 /Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pyld/jsonld.py:775(expand)
 2921/150    0.109    0.000   63.181    0.421 /Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pyld/jsonld.py:2162(_expand)
43563054/12068   28.329    0.000   59.429    0.005 /Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/copy.py:132(deepcopy)
...
```

Of 63 seconds, 59 of them are spent in deepcopy()

gkellogg:
We should see how important those copies actually are. In some cases, I think they're because of the possibility of things being modified; If this were Virtual Memory, we'd do some form of copy-on-write operation.

I'll look over it more tomorrow.

azaroth42:
But indeed the codepaths are not from this PR. The same profiling task on the `1.1-expansion` branch is much much worse again, by another factor of 10:

```
      150    0.001    0.000  893.278    5.955 /Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pyld/jsonld.py:129(expand)
      150    0.023    0.000  893.277    5.955 /Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pyld/jsonld.py:764(expand)
 1900/150    0.098    0.000  893.238    5.955 /Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pyld/jsonld.py:2133(_expand)
45265/823    3.079    0.000  888.235    1.079 /Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pyld/jsonld.py:2953(_process_context)
501207/371996    4.545    0.000  850.570    0.002 /Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pyld/jsonld.py:4854(_create_term_definition)
  732/150    0.027    0.000  708.161    4.721 /Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pyld/jsonld.py:2363(_expand_object)
531008400/83636  339.041    0.000  690.415    0.008 /Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/copy.py:132(deepcopy)
```

gkellogg:
@azaroth. Perhaps you can share some data and scripts you’re using for doing the benchmark. I suspect that more use of frozen objects can help to eliminate deep copies altogether.

azaroth42:
Here's my profiler code, context and 150 example dataset:
[profiler.zip](https://github.com/digitalbazaar/pyld/files/4424796/profiler.zip)

azaroth42:
@gkellogg Would you like us to test this branch (or another one?) against the full dataset? Or how do we feel confident to merge into 1.1, and then through to master?

gkellogg:
@azaroth42 this branch has all the latest changes for caching and should give the best performance. Hopefully, this and other branches will be merged into master sooner rather than later (particularly, as we need that to get an EARL report out for the WG implementation report).

azaroth42:
@kasei reports this is 
> "WAY FASTER"  
But needs the setup to include the `frozendict` module.  (Which I noted when I was running it, but was too curious to see the results to pay attention to the details :D)

We're running our processing pipeline to compare.

gkellogg:
> @kasei reports this is
> 
> > "WAY FASTER"
> 
> But needs the setup to include the `frozendict` module. (Which I noted when I was running it, but was too curious to see the results to pay attention to the details :D)
> 
> We're running our processing pipeline to compare.

I added both 'frozendict' and 'uuid' to setup.py.

azaroth42:
Details about "way faster" ... this branch as of a82c753 does expansion of our dataset in 4h20, whereas master takes 9h34 for exactly the same set of documents :)

gkellogg:
So, a bit more than twice as fast. Is that fast enough?

azaroth42:
Yep it's great for us, especially as master doesn't have all of the other
functionality that you've added in this and the 1.1 branch!

PR #113, opened 2020-04-03
Immutable contexts

gkellogg:
Reduce deep copies by freezing the context and specifically enabling places that would have modified it after generation.

cc/ @azaroth42

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ cover
 dist
 docs/_build
 lib/PyLD.egg-info
+profiler
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ matrix:
       env: LOADER=aiohttp
   allow_failures:
     - python: "3.4"
+    - python: "3.5"
 
 install:
   - pip install -r requirements.txt
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@
   * Don't set `@base` in initial context and don't resolve a relative IRI
     when setting `@base` in a context, so that the document location can
     be kept separate from the context itself.
+  * Use static initial contexts composed of just `mappings` and `processingMode`
+    to enhance preprocessed context cachability.
 * Create Term Definition:
   * Allow `@type` as a term under certain circumstances.
   * Reject and warn on keyword-like terms.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@
   * Added a ContextResolver which can used a shared LRU cache for storing
     externally retrieved contexts, and the result of processing them relative
     to a particular active context.
+  * Return a "frozendict" from context processing and reduce deepcopies.
+  * Store inverse context in an LRU cache rather than trying to modify a frozen context.
 * Create Term Definition:
   * Allow `@type` as a term under certain circumstances.
   * Reject and warn on keyword-like terms.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # pyld ChangeLog
 
 ### Changed
-* **BREAKING**: Versions of Python before 3.4 are no longer supported.
+* **BREAKING**: Versions of Python before 3.6 are no longer supported.
 * Update conformance docs.
 * Add all keywords and update options.
 * Default processingMode to json-ld-1.1.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@
     to a particular active context.
   * Return a "frozendict" from context processing and reduce deepcopies.
   * Store inverse context in an LRU cache rather than trying to modify a frozen context.
+  * Don't set `@base` in initial context and don't resolve a relative IRI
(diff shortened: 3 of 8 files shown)


gkellogg:
@azaroth42 It's better, but a bunch of calls to frozendict and _create_term_definition, which I wouldn't expect to much when reusing a context. There's still some work to do, but it seems faster.

```
         27991557 function calls (19130548 primitive calls) in 10.582 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      150    0.001    0.000   10.582    0.071 profiler/../lib/pyld/jsonld.py:140(expand)
      150    0.001    0.000   10.581    0.071 profiler/../lib/pyld/jsonld.py:777(expand)
 2921/150    0.016    0.000   10.565    0.070 profiler/../lib/pyld/jsonld.py:2164(_expand)
9024/1127    0.160    0.000   10.361    0.009 profiler/../lib/pyld/jsonld.py:2984(_process_context)
7256731/9003    3.132    0.000    6.892    0.001 profiler/../lib/pyld/jsonld.py:6389(freeze)
1505009/9003    1.322    0.000    6.869    0.001 profiler/../lib/pyld/jsonld.py:6391(<listcomp>)
    93555    0.709    0.000    2.433    0.000 profiler/../lib/pyld/jsonld.py:4856(_create_term_definition)
  1524516    0.954    0.000    0.954    0.000 {method 'items' of 'dict' objects}
  9379263    0.890    0.000    0.919    0.000 {built-in method builtins.isinstance}
  1505009    0.834    0.000    0.834    0.000 /usr/local/lib/python3.7/site-packages/frozendict/__init__.py:24(__init__)
 1109/150    0.021    0.000    0.759    0.005 profiler/../lib/pyld/jsonld.py:2394(_expand_object)
   489322    0.175    0.000    0.691    0.000 /usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/re.py:170(match)
   109030    0.238    0.000    0.583    0.000 profiler/../lib/pyld/jsonld.py:5291(_expand_iri)
```

gkellogg:
It turns out that the initial context always has `@base` set to the document location, and as this is used to index into the ProcessedContext cache, it always end up missing.

The solution would seem to be to take `@base` out of that context (just leaving `processingMode`) and figure out how to do that at the end of `_process_context` using the options value if the resulting context doesn't have an `@base` element. This does mean returning a modified context, but should avoid other processing.

@dlongley jsonld.js has the same issue.

I also note that trying to do a cache of `_get_initial_context` introduces some instability, with some tests passing or failing intermittently. As it is, there are some periodic failures in the travis tests. My guess is that using `id()` on the active context is unreliable, and it sometimes hits on a context which was cached from a previous file. This shouldn't happen, but I don't know what else to tag it too. The solution may be, as is done in jsonld.js, to stringily the active context to use as a cache key, or a hash of that stringification, but you wouldn't think this would be necessary.

azaroth42:
I get similar profiling results, unsurprisingly :)

gkellogg:
Hmm, the test ran substantially faster for me, with much less time in deepcopy. But, `_process_context` is still up at the top.

azaroth42:
Apologies, I meant similar results *to yours*, rather than to my previous results :)

gkellogg:
@azaroth42 still dealing with key management fo context caching, but this update now does not put the base IRI in the active context, and it seems to hit the cache much better.

```
         3737480 function calls (3118485 primitive calls) in 1.099 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      150    0.000    0.000    1.099    0.007 profiler/../lib/pyld/jsonld.py:143(expand)
      150    0.001    0.000    1.098    0.007 profiler/../lib/pyld/jsonld.py:780(expand)
 2921/150    0.015    0.000    1.083    0.007 profiler/../lib/pyld/jsonld.py:2168(_expand)
1180/1127    0.018    0.000    0.877    0.001 profiler/../lib/pyld/jsonld.py:2990(_process_context)
 1109/150    0.021    0.000    0.751    0.005 profiler/../lib/pyld/jsonld.py:2398(_expand_object)
496829/998    0.121    0.000    0.295    0.000 profiler/../lib/pyld/jsonld.py:6408(freeze)
14731/998    0.080    0.000    0.292    0.000 profiler/../lib/pyld/jsonld.py:6410(<listcomp>)
     1974    0.118    0.000    0.213    0.000 profiler/../lib/pyld/jsonld.py:5486(_clone_active_context)
1330/1180    0.004    0.000    0.176    0.000 profiler/../lib/pyld/context_resolver.py:30(resolve)
  1122009    0.141    0.000    0.169    0.000 {built-in method builtins.isinstance}
```

gkellogg:
With change to use UUIDs, seems to work consistently for Python 3.6 and greater.

gkellogg:
Since the base context-resolver branch has issues anyway, I'm going to go ahead and merge this into that.

There are things here to consider applying to jsonld.js, particularly removing `base` from the initial context.

PR #120, opened 2020-04-17
Fix inverse context cache indexing to use the uuid field

### LINKED ISSUE #119: Issue when repeatedly expanding and compacting jsonld with nested/type-scoped contexts
Trying out PyLD 2.0.1 (Also happens with PyLD 2.0.0) on Python 3.6-3.8 and everything works great, performance improves significantly, but tests in our test suite started failing randomly.

After a lot of trial and error, it seems like when expanding and compacting, at random, contexts aren't applied correctly.

For instance 
```
{
"@context": { 
    "schema": "http://schema.org/", 
    "name": "schema:name"
},
"name": "test"
}
```

ends up being 

```
{
"@context": { 
    "schema": "http://schema.org/", 
    "name": "schema:name"
},
"schema:name": "test"
}
```

which breaks our processing down the line.

I made a minimal example that showcases the issue. 
[pyld_reproduce_error.zip](https://github.com/digitalbazaar/pyld/files/4491421/pyld_reproduce_error.zip)

The example loads the attached jsonld file, expands it and compacts it again with the original context. This is repeated until the data after expansion&compaction does not match the output of the last iteration of the loop.

The problem crops up anywhere between iteration 1 and 32 (average 6.5) when I ran it 100 times, it seems completely random when it crops up (Which in our test suite caused tests to fail at random).

Any idea what could be causing this?

gkellogg:
instead of the object id of the active context.

Fixes #119.

@Panaetius, please see if using this branch solves the problem for you.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ dist
 docs/_build
 lib/PyLD.egg-info
 profiler
+tests/test_caching.py
+tests/data/test_caching.json
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # pyld ChangeLog
 
+### Changed
+* Fix inverse context cache indexing to use the uuid field
+
 ## 2.0.1 - 2020-04-15
 
 ### Changed
--- a/lib/pyld/jsonld.py
+++ b/lib/pyld/jsonld.py
@@ -5540,7 +5540,7 @@ class JsonLdProcessor(object):
         :return: the inverse context.
         """
         # inverse context already generated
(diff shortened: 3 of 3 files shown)


gkellogg:
The problem was due to code which was not replaced which used `id(active_ctx)` rather than `active_ctx['_uuid']` to index into the inverse context cache, and it seems that object identifiers are recycled.

Panaetius:
@gkellogg Thank you very much for the quick reply and fix! Yes this does solve the issue for me and all tests on our side pass when using this branch.

davidlehn:
Released in 2.0.2.

PR #131, opened 2020-07-16
Remove multiple header entries in call to requests.get()

### LINKED ISSUE #125: get() got multiple values for keyword argument 'headers'
There is a problem with `requests_document_loader`. When using custom headers like this `jsonld.set_document_loader(jsonld.requests_document_loader(headers=...))` the `headers=` argument is [repeated twice (here)](https://github.com/digitalbazaar/pyld/blob/master/lib/pyld/documentloader/requests.py#L63) and Python raises an error.

cgils:
This suggested fix adds headers to ```kwargs``` in ```documentloader.requests.loader()```. Feel free to not accept and close if there's a better way of doing this. AFAICT, mutating ```kwargs``` locally should be safe.

Fixes the problem described in issue #125 when supplying user-defined headers to ```jsonld.requests_document_loader```:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/pyld/documentloader/requests.py", line 64, in loader
    response = requests.get(url, headers=headers, **kwargs)
TypeError: get() got multiple values for keyword argument 'headers'
```

Closes #125

### DIFF
--- a/lib/pyld/documentloader/requests.py
+++ b/lib/pyld/documentloader/requests.py
@@ -60,7 +60,8 @@ def requests_document_loader(secure=False, **kwargs):
                 headers = {
                     'Accept': 'application/ld+json, application/json'
                 }
-            response = requests.get(url, headers=headers, **kwargs)
+                kwargs['headers'] = headers
+            response = requests.get(url, **kwargs)
 
             content_type = response.headers.get('content-type')
             if not content_type:


lib/pyld/documentloader/requests.py
```diff
@@ -60,7 +60,8 @@ def loader(url, options={}):
                 headers = {
                     'Accept': 'application/ld+json, application/json'
                 }
-            response = requests.get(url, headers=headers, **kwargs)
+                kwargs['headers'] = headers
```
sharpaper, line 63: I don't think this patch will fix the issue. From what I understand there are 2 places where headers are retrieved. One is the "options" parameter in `def loader(url, options={})` and the other one is from the "kwargs" in `def requests_document_loader(secure=False, **kwargs)`. kwargs is used to pass user arguments to `requests`, and adding `                kwargs['headers'] = headers` will simply remove/replace any headers. It should probably be a merge of the "headers" speficied in `def loader(url, options={})` and `def requests_document_loader(secure=False, **kwargs)`.

cgils:
Closing in favor of a more appropriate fix.

PR #135, opened 2020-08-03
Issue 128

datadavev:
Addresses issue #128 for the requests document downloader. This adds an optional keyword parameter `max_link_follows` (defaults to 2) to `requests_document_loader` to avoid possible infinite recursion following link headers.


datadavev:
Retracting in favor of https://github.com/digitalbazaar/pyld/pull/129

Issue #137, opened 2020-09-14
to_rdf silently ignores invalid IRIs


aucampia:
When I run `to_rdf` on an object with an invalid `@id`:

```python
import pyld.jsonld as jsonld
result = jsonld.to_rdf({"@id": "http://example.com/some id", "@type": "http://example.com/some_type"})
```
The the value of `result` will be
```
{'@default': []}
```

In contrast, if I fix the `@id` to be a valid URI

```python
import pyld.jsonld as jsonld
result = jsonld.to_rdf({"@id": "http://example.com/some_id", "@type": "http://example.com/some_type"})
```
The the value of `result` will be
```
{'@default': [{'subject': {'type': 'IRI', 'value': 'http://example.com/some_id'}, 'predicate': {'type': 'IRI', 'value': 'http://www.w3.org/1999/02/22-rdf-syntax-ns#type'}, 'object': {'type': 'IRI', 'value': 'http://example.com/some_type'}}]}
```

I expect that running `to_rdf` on JSON-LD with invalid IRIs error out instead of silently ignoring the invalid nodes.

aucampia:
I think the right solution is to change https://github.com/digitalbazaar/pyld/blob/316fbc2c9e25b3cf718b4ee189012a64b91f17e7/lib/pyld/jsonld.py#L3551 from `continue` to `raise ...`.

mielvds:
@aucampia in case you're still interested: I think this is expected behaviour. According to the [Object to RDF Conversion in the spec](https://www.w3.org/TR/json-ld11-api/#object-to-rdf-conversion):

> If item is a [node object](https://www.w3.org/TR/json-ld11/#dfn-node-object) and the value of its @id [entry](https://infra.spec.whatwg.org/#map-entry) is not [well-formed](https://www.w3.org/TR/json-ld11-api/#dfn-well-formed), return null.

and

> If a [node object](https://www.w3.org/TR/json-ld11/#dfn-node-object) containing a [relative IRI reference](https://tools.ietf.org/html/rfc3987#section-6.5) is passed to the algorithm, null is returned which then causes the resulting [triple](https://www.w3.org/TR/rdf11-concepts/#dfn-rdf-triple) to be ignored.

PR #138, opened 2020-09-14
Treat invalid IRIs as errors in conversion to RDF

aucampia:
Currently invalid (i.e. non absolute) IRIs are ignored and tripples
containing non absolute IRIs are omitted from the RDF resulting from
`to_rdf`.

This patch changes this behaviour so that non absolute IRIs result in
exceptions instead.

See https://github.com/digitalbazaar/pyld/issues/137

I am not sure what to do about testing, it seems there are no general unit tests, but rather just some scaffolding to run the w3c tests (which fail).

If you want I can add some general unit tests for you, just need to agree on the framework I guess.

### DIFF
--- a/lib/pyld/jsonld.py
+++ b/lib/pyld/jsonld.py
@@ -3545,10 +3545,16 @@ class JsonLdProcessor(object):
                     continue
 
                 for item in items:
-                    # skip relative IRI subjects and predicates
+                    # All IDs should be absolute at this point in time
                     if not (_is_absolute_iri(id_) and
                             _is_absolute_iri(property)):
-                        continue
+                        raise JsonLdError(
+                            'JSON-LD graph conversion error; '
+                            'invalid IRI for property or ID, '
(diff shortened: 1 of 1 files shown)


gkellogg:
The spec specifically calls for such IRIs to be ignored, so raising an error would cause some tests to fail. I think the best thing to do is issue a warning, but a processor flag could cause an error to be raised.

aucampia:
Thanks for the info @gkellogg 

I think maybe to keep things clean it is best to add two options: `strict` and `warn`. If `strict` is set to `True` then raise when encountering an invalid IRI (i.e. not a valid absolute IRI), if `warn` is set to `True` then warn in the same case.

`strict` and `warn` could then be used for other similar cases in future.

Not sure what you think is best, feedback is appreciated.

aucampia:
The other option is maybe to just have `warn` and `errorOnInvalidIRI` - but I worry that this will lead to proliferation of warning options.

Issue #133, opened 2020-08-01
Could not retrieve a JSON-LD document from the URL.


sharpaper:
For some reasons pyld is not able to fetch the context URL `https://www.w3.org/ns/activitystreams`. I was able to reduce the issue to the following basic example, which is not working:

    import pyld
    doc = {
        "@context": [ 'https://www.w3.org/ns/activitystreams' ],
        "type": "Follow" }
    doc = pyld.jsonld.expand(doc)

    [...]
    raise JSONDecodeError("Expecting value", s, err.value) from None
    [...]
    Dereferencing a URL did not result in a valid JSON-LD object.
    [...]
    Type: jsonld.InvalidUrl
    Code: loading remote context failed
    Details: {'url': 'https://www.w3.org/ns/activitystreams', 'cause': JsonLdError('Could not retrieve a JSON-LD document from the URL.')}

I could not debug the issue. It looks like pyld might be retrieving the context without the correct headers, but looking at the code I do see `headers = { 'Accept': ...}` defined in several places.
Can you guys please help me understand if this is a bug or if I'm not using the library correctly? Thanks!

sharpaper:
If this can be any useful, when the "document loader" sends the `.get()` request to retrieve the remote context, see [L63](https://github.com/digitalbazaar/pyld/blob/master/lib/pyld/documentloader/requests.py#L63), the parameters are:

    url = https://www.w3.org/ns/activitystreams
    headers = {'Accept': 'application/ld+json;profile=http://www.w3.org/ns/json-ld#context, application/ld+json, application/json;q=0.5, text/html;q=0.8, application/xhtml+xml;q=0.8'}
    kwargs = {}

Because the request accepts everything, the remote has selected HTML and therefore pyld gives the error. `curl`-ing the URI with `'Accept': 'application/ld+json;profile=http://www.w3.org/ns/json-ld#context, application/ld+json, application/json;q=0.5'`, after removing the HTML option from Accept, does work.
So I think the website might be at fault here because `application/ld+json` should take higher precedence over `text/html;q=0.8`, but on the other hand why is pyld requesting HTML? Can HTML be removed from the Accept headers somehow?

sharpaper:
[...]

PR #144, opened 2020-12-19
avoid duplicate "header" kwarg

cedricvanrompay:
Fixes https://github.com/digitalbazaar/pyld/issues/133

If headers were passed during call to `requests_document_loader`,
these headers will be forced in call to `requests.get`.

### DIFF
--- a/lib/pyld/documentloader/requests.py
+++ b/lib/pyld/documentloader/requests.py
@@ -14,6 +14,9 @@ import urllib.parse as urllib_parse
 
 from pyld.jsonld import (JsonLdError, parse_link_header, LINK_HEADER_REL)
 
+DEFAULT_HEADERS = {
+    'Accept': 'application/ld+json, application/json'
+}
 
 def requests_document_loader(secure=False, **kwargs):
     """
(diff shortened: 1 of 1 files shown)


lib/pyld/documentloader/requests.py
```diff
@@ -55,12 +58,16 @@ def loader(url, options={}):
                     'the URL\'s scheme is not "https".',
                     'jsonld.InvalidUrl', {'url': url},
                     code='loading document failed')
-            headers = options.get('headers')
-            if headers is None:
-                headers = {
-                    'Accept': 'application/ld+json, application/json'
-                }
-            response = requests.get(url, headers=headers, **kwargs)
```
anatoly-scherbakov, line 63: It is unclear how this PR is changing the behavior. From what I am seeing, even before the PR a user could do

```python
requests_document_loader(…, headers={
    'Foo': 'bar',
})
```

it should be working on implementation at `master`. Or am I misunderstanding it?

lib/pyld/documentloader/requests.py
```diff
@@ -55,12 +58,16 @@ def loader(url, options={}):
...
+            
+            headers = {} if 'headers' in kwargs else ({
+                'headers': (
+                    options.get('headers', None)
```
anatoly-scherbakov, line 64: Why we're addressing both `kwargs['headers']` and `options['headers']`, where should those be found?

lib/pyld/documentloader/requests.py
```diff
@@ -14,6 +14,9 @@
 
 from pyld.jsonld import (JsonLdError, parse_link_header, LINK_HEADER_REL)
 
+DEFAULT_HEADERS = {
+    'Accept': 'application/ld+json, application/json'
+}
```
anatoly-scherbakov, line 19: :+1: I appreciate factoring this out as a global constant.

BigBlueHat:
Pretty sure this just avoids the code path where the HTML media types are added. Also, #170 has some fixes related to #133 and I believe that's now been fixed upstream by @pchampin and/or @iherman

So, gonna close this one in favor of those.

PR #153, opened 2021-05-14
ability to fetch from local ipfs instance

drbh:
not sure if this functionality is on the roadmap but reading context's from IPFS are helpful to a personal use case.

Please let me know what you think, and what changes would be needed to add this to the library

Use example
```python
from pyld import jsonld
import json

context = {
  "@context": {
    "name": "http://schema.org/name",
    "image": {
      "@id": "http://schema.org/image",
      "@type": "@id"
    },
    "homepage": {
      "@id": "http://schema.org/url",
      "@type": "@id"
    }
  }
}

import ipfshttpclient
client = ipfshttpclient.connect()
cid = client.add_json(context)

doc = {
  "@context": f"ipfs://{cid}",
  "name": "Manu Sporny",
  "homepage": "http://manu.sporny.org/",
  "image": "http://manu.sporny.org/images/manu.png"
}

compacted = jsonld.compact(doc, context)

# {
#   "@context": {
#     "name": "http://schema.org/name",
#     "image": {
#       "@id": "http://schema.org/image",
#       "@type": "@id"
#     },
#     "homepage": {
#       "@id": "http://schema.org/url",
#       "@type": "@id"
#     }
#   },
#   "image": "http://manu.sporny.org/images/manu.png",
#   "name": "Manu Sporny",
#   "homepage": "http://manu.sporny.org/"
# }
```

### DIFF
--- a/lib/pyld/documentloader/requests.py
+++ b/lib/pyld/documentloader/requests.py
@@ -41,11 +41,11 @@ def requests_document_loader(secure=False, **kwargs):
             # validate URL
             pieces = urllib_parse.urlparse(url)
             if (not all([pieces.scheme, pieces.netloc]) or
-                pieces.scheme not in ['http', 'https'] or
+                pieces.scheme not in ['http', 'https', 'ipfs'] or
                 set(pieces.netloc) > set(
                     string.ascii_letters + string.digits + '-.:')):
                 raise JsonLdError(
-                    'URL could not be dereferenced; only "http" and "https" '
+                    'URL could not be dereferenced; only "http", "https" and "ipfs" '
                     'URLs are supported.',
                     'jsonld.InvalidUrl', {'url': url},
                     code='loading document failed')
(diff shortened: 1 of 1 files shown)


anatoly-scherbakov:
@drbh this might be outdated, but I am also interested in using IPFS with `pyld`. I would believe though that it would be beneficial to create a separate Document Loader for IPFS instead of adding an `if` to existing `requests` based loader. Did you pursue that or a different venue to achieve your goal? Were you successful? Is there any public info about your project?

lib/pyld/documentloader/requests.py
```diff
@@ -60,7 +60,22 @@ def loader(url, options={}):
                 headers = {
                     'Accept': 'application/ld+json, application/json'
                 }
-            response = requests.get(url, headers=headers, **kwargs)
+
+            if pieces.scheme == "ipfs":
```
anatoly-scherbakov, line 64: As I have already commented earlier, this should probably be refactored to a separate project.

anatoly-scherbakov (changes requested):

BigBlueHat:
The jsonld.js library has pluggable document loaders--which looks to be what's actually needed here vs. adding custom loader code directly: https://github.com/digitalbazaar/jsonld.js#custom-document-loader

I'd suggest exploring implementing something like that custom document loader approach and then refactoring this to use that approach for pluggability.

drbh:
closing since this is a bit outdated and a better impl strategy has been suggested above. Thanks!

Issue #156, opened 2021-05-26
CONTRIBUTING.rst needs to be revised to reflect W3C test repository


hsolbrig:
CONTRIBUTING.rst  states:
```
As of early 2020, the process to generate an EARL report for the official JSON-LD Processor Conformance page is:

    Run the tests on the json-ld-api and json-ld-framing test repos to generate a .jsonld test report:
        python tests/runtests.py ../json-ld-api/tests/ ../json-ld-framing/tests/ -e pyld-earl.jsonld
```
which sort of implies that we should be using the `digitalbazaar` tests.  We should update this to make it clear that these tests now come from the w3c repositories

hsolbrig:
Note: I'll submit a pull request for this fix

PR #157, opened 2021-05-26
Clarify documentation on runing Earl reports

hsolbrig:
fixes issue #156

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -44,14 +44,23 @@ Implementation Report Process
 As of early 2020, the process to generate an EARL report for the official
 `JSON-LD Processor Conformance`_ page is:
 
+* Check out a local copy of the test suites::
+
+    > mkdir ../w3c
+    > cd ../w3c
+    > git clone https://github.com/w3c/json-ld-api.git
+    > git clone https://github.com/w3c/json-ld-framing.git
+    > cd ../pyld
+
+
 * Run the tests on the ``json-ld-api`` and ``json-ld-framing`` test repos to
   generate a ``.jsonld`` test report:
 
-  * ``python tests/runtests.py ../json-ld-api/tests/ ../json-ld-framing/tests/ -e pyld-earl.jsonld``
+  > ``python tests/runtests.py ..w3c/json-ld-api/tests/ ../w3c/json-ld-framing/tests/ -e pyld-earl.jsonld``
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -44,14 +44,23 @@ Implementation Report Process
 As of early 2020, the process to generate an EARL report for the official
 `JSON-LD Processor Conformance`_ page is:
 
+* Check out a local copy of the test suites::
```
davidlehn, line 47: I think it might be better to align with the main README test section for this: https://github.com/digitalbazaar/pyld#tests.  The opinionated setup there and in runtests.py is to checkout the test dirs as siblings of pyld.  runtests.py defaults to looking for those dirs so you don't have to list them on command line.  In the case of EARL report generation, the test dirs are explicit to avoid also running the normalization tests.  As is here you'd be instructed to check out dirs in different locations than when running regular tests.  Thoughts?

davidlehn: Maybe this could just refer to the readme test section for setup?

CONTRIBUTING.rst
```diff
@@ -44,14 +44,23 @@ Implementation Report Process
...
+    > git clone https://github.com/w3c/json-ld-api.git
+    > git clone https://github.com/w3c/json-ld-framing.git
+    > cd ../pyld
+
+
 * Run the tests on the ``json-ld-api`` and ``json-ld-framing`` test repos to
   generate a ``.jsonld`` test report:
 
-  * ``python tests/runtests.py ../json-ld-api/tests/ ../json-ld-framing/tests/ -e pyld-earl.jsonld``
+  > ``python tests/runtests.py ..w3c/json-ld-api/tests/ ../w3c/json-ld-framing/tests/ -e pyld-earl.jsonld``
```
davidlehn, line 59: As stated above, I'm not sure this w3c subdir idea is the best approach.  But if so, first path is missing a `/`.

hsolbrig: I'll leave it up to you how it should be resolved -- mostly it is just that I went directly to CONTRIBUTING and there wasn't anything that prevented me from interpreting it as asking for the tests to be run against the digitalbazaar tests instead of the W3C.

BigBlueHat:
Closing in favor of #182

davidlehn:
@BigBlueHat I don't think #182 addresses #156 as was the intent here.  But it does change what the docs might say.  I'll note it there.

Issue #163, opened 2022-06-29
IndexError in jsonld.py_value_match


ioggstream:
## Reproduce

when JsonLdProcessor.get_values(pattern, "@value") == []

## I expect

return False

## Instead

IndexError

PR #164, opened 2022-06-29
Fix: #163. IndexError when matching values.

ioggstream:
## This PR

- checks a value before accessing an index.

### DIFF
--- a/lib/pyld/jsonld.py
+++ b/lib/pyld/jsonld.py
@@ -4504,7 +4504,7 @@ class JsonLdProcessor(object):
 
         if not v2 and not t2 and not l2:
             return True
-        if not (v1 in v2 or _is_empty_object(v2[0])):
+        if not (v1 in v2 or not v2 or _is_empty_object(v2[0])):
             return False
         if not ((not t1 and not t2) or (t1 in t2) or (t1 and t2 and _is_empty_object(t2[0]))):
             return False


ioggstream:
cc: @dlongley @msporny is this still maintained?

msporny:
> cc: @dlongley @msporny is this still maintained?

Barely... it's not a priority for us because we no longer use Python. We've been searching for a maintainer that could take over maintenance of this library for several years now (to no avail).

We do plan on setting aside some money for a maintainer in the coming year... do you know of anyone that would be interested in maintaining the library? Our engineering team is too busy in other areas at the moment (VCs, DIDs, Data Integrity, etc... all building on top of JSON-LD 1.1).

dlongley (approved):
LGTM.

dlongley (changes requested):
On second thought, I'm not sure this is right when comparing it to another implementation here:

https://github.com/digitalbazaar/jsonld.js/blob/main/lib/frame.js#L813

In python, the statement `not v2`, when `v2` is an empty array (which seems to be the case of concern here) results in `True`. But we are looking to return `False` only when `v1` is actually in `v2` or when `v2` has a first element that is an empty object. This would erroneously return `False` when `v2` is an empty array or isn't present (because it defaults to an empty array).

lib/pyld/jsonld.py
```diff
@@ -4504,7 +4504,7 @@ def _value_match(self, pattern, value):
 
         if not v2 and not t2 and not l2:
             return True
-        if not (v1 in v2 or _is_empty_object(v2[0])):
+        if not (v1 in v2 or not v2 or _is_empty_object(v2[0])):
```
dlongley, line 4507: This should probably be:

```suggestion
        if not (v1 in v2 or (v2 and _is_empty_object(v2[0]))):
```

Ideally tests would confirm.

ioggstream:
Seems reasonable. I'll provide further comment in the code though.

ioggstream:
@mielvds thanks! I forgot it, sorry :(

mielvds:
> @mielvds thanks! I forgot it, sorry :(

makes sense; this is from '22 😄 Let's see if there are any tests failing

PR #170, opened 2022-11-20
load_document: Lower priority of HTML inputs

cid-chan:
This fixes a bug with activitypub related documents: The schema "https://www.w3.org/ns/activitystreams" renders the spec if "application/json" has a lower prioritisation than "text/html".

### DIFF
--- a/lib/pyld/jsonld.py
+++ b/lib/pyld/jsonld.py
@@ -6334,11 +6334,18 @@ def load_document(url,
 
     :return: True if the value is an absolute IRI, False if not.
     """
+    # Prefer JSON-LD and JSON (q=0.8) over HTML (q=0.5) to ensure proper
+    # content negotiation with servers that serve both formats at the same URL,
+    # such as https://www.w3.org/ns/activitystreams
+    # See: https://github.com/digitalbazaar/pyld/pull/170
     headers = {
-        'Accept': 'application/ld+json, application/json;q=0.8'
+        'Accept': (
+            'application/ld+json, '
+            'application/json;q=0.8, '
+            'text/html;q=0.5, '
+            'application/xhtml+xml;q=0.5'
+        )
     }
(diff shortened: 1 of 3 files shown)


lib/pyld/jsonld.py
```diff
@@ -6568,10 +6568,10 @@ def load_document(url,
     :return: True if the value is an absolute IRI, False if not.
     """
     headers = {
-        'Accept': 'application/ld+json, application/json;q=0.5'
+        'Accept': 'application/ld+json, application/json;q=0.8'
```
anatoly-scherbakov, line 6571: We define `headers['Accept']` here…

lib/pyld/jsonld.py
```diff
@@ -6568,10 +6568,10 @@ def load_document(url,
...
     }
     # FIXME: only if html5lib loaded?
-    headers['Accept'] = headers['Accept'] + ', text/html;q=0.8, application/xhtml+xml;q=0.8'
+    headers['Accept'] = headers['Accept'] + ', text/html;q=0.5, application/xhtml+xml;q=0.5'
```
anatoly-scherbakov, line 6574: …and we modify that value here. 

* Would it be possible to specify the final value right away?
* Probably a comment explaining what the magic `0.5` means would also be helpful to future readers.

anatoly-scherbakov (changes requested):

BigBlueHat:
This looks fine generally, but lets get some tests written for it. @anatoly-scherbakov would you be up for doing that to test out the "local tests" you were sad we didn't have here?

anatoly-scherbakov (approved):

mielvds (approved):
Looks good! The only thought I had was about the use of class-based vs function-based tests. OK for merging as is, but I might merge this with another test file in #228 later on.

PR #185, opened 2023-11-03
Switch from travis-ci to GitHub workflow.

davidlehn:


.github/workflows/main.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - uses: actions/checkout@v4
+    - name: Use Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
+        cache: 'pip'
+    - name: Install testing dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements-test.txt
```
anatoly-scherbakov, line 23: I thought `requrements.txt` already contained test dependencies, so could we just add the new packages to that file perhaps?

.github/workflows/main.yaml
```diff
@@ -0,0 +1,73 @@
...
+      env:
+        LOADER: ${{ matrix.loader }}
+  test:
+    runs-on: ubuntu-latest
+    needs: [lint]
+    timeout-minutes: 10
+    strategy:
+      matrix:
+        python-version:
+          - '3.8'
```
anatoly-scherbakov, line 38: I :clap: applaud the decision to deprecate currently unsupported Python versions, which are < 3.8.

.github/workflows/main.yaml
```diff
@@ -0,0 +1,73 @@
...
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+    - name: Install testing dependencies
+      run: |
+        pip install -r requirements-test.txt
+    - name: Fetch test suites
+      run: |
+        git clone --depth 1 https://github.com/w3c/json-ld-api.git _json-ld-api
+        git clone --depth 1 https://github.com/w3c/json-ld-framing.git _json-ld-framing
+        git clone --depth 1 https://github.com/json-ld/normalization.git _normalization
```
anatoly-scherbakov, line 63: These three aren't going to be necessary if https://github.com/digitalbazaar/pyld/pull/182 is merged.

.github/workflows/main.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - name: Test with Python=${{ matrix.python-version }} Loader=${{ matrix.loader }}
+      run: |
+        python tests/runtests.py ./_json-ld-api/tests -l ${{ matrix.loader }}
+        python tests/runtests.py ./_json-ld-framing/tests -l ${{ matrix.loader }}
+        python tests/runtests.py ./_normalization/tests -l ${{ matrix.loader }}
```
anatoly-scherbakov, line 68:
```suggestion
        python tests/runtests.py -l ${{ matrix.loader }}
```

is going to be sufficient.

anatoly-scherbakov (changes requested):

.github/workflows/main.yaml
BigBlueHat: @anatoly-scherbakov the `submodule` command will have to be added.

.github/workflows/main.yaml
BigBlueHat: `flake8` isn't in `requirements.txt`, so I think this is till needed for the additional test code below.

davidlehn: The intent of `requiremenst-test.txt` was to avoid those deps being installed other than for testing.  No need to install a linter for a regular install.  What is best practice for that concept?  I saw that somewhere and it seemed a reasonable approach.

.github/workflows/main.yaml
davidlehn: - I believe what I was thinking when this PR was written was to start with testing the code similar to how travis-ci was doing it but with github actions.
- This will give immediate familiar feedback for testing further changes.
- #182 is not addressing the test runner and code changes needed for rdf-canon tests.  Every rdf-canon test will be skipped leaving a large hole in coverage if we are to release any versions before the issue is addressed.  I don't think changing that test suite before rdf-canon is supported can be justified.
- It's arguably better to merge this *first* to get tests flowing.  #182 can be rebased off of it and will then have tests running.  The changes to make it use submodules are probably straightforward and would be best done with workflow tests to confirm.  (Assuming we can get outside contributor PR tests running properly.)

anatoly-scherbakov: > is not addressing the test runner and code changes needed for rdf-canon tests. Every rdf-canon test will be skipped

Should we fix that? I thought I [did add](https://github.com/digitalbazaar/pyld/pull/182/files#diff-fe7afb5c9c916e521401d3fcfb4277d5071798c3baf83baf11d6071742823584R9) that as a dependency and we should be running against that there...

.github/workflows/main.yaml
```diff
@@ -0,0 +1,73 @@
...
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    strategy:
+      matrix:
+        python-version:
+          - '3.12'
+    steps:
+    - uses: actions/checkout@v4
```
anatoly-scherbakov, line 14:
```suggestion
    - uses: actions/checkout@v4
      with:
        submodules: true
```

As per https://github.com/actions/checkout

Applies to the below again.

.github/workflows/main.yaml
anatoly-scherbakov: @BigBlueHat it seems that `actions/checkout` has an option to fetch submodules so we can likely avoid running `make` even.

anatoly-scherbakov (changes requested):

BigBlueHat:
I'm in favor of getting this merged sooner than later--and it can happen now (vs. as part of a release) since it only affects the `master` branch and nothing about the running code (or even the tests themselves).

@anatoly-scherbakov we should separate the `rdf-canon` introduction from #182, so both can move forward at different speeds.

.github/workflows/main.yaml
BigBlueHat: Good to know we can do this, but I agree with @davidlehn that we should get this merged before #182 and can revisit this if/when #182 makes it in.

ff137:
May I ask what is the current state of the project?
Last comment says merging can happen now, but that's more than a year ago

davidlehn:
- Trying to update this change to it can be merged.
- Too many other issues getting in the way, they can be handled on top of this with other PRs.
- This was hanging out so long that 2 python versions hit EOL (or are set to this month).
- 3.14 has a new event loop issue that causes aiohttp to fail.  Anyone know a small non-invasive fix?

.github/workflows/main.yaml
mielvds: @anatoly-scherbakov @davidlehn let's discuss the use of rdf-canon in https://github.com/digitalbazaar/pyld/issues/190 and resolve this conversation.

.github/workflows/main.yaml
mielvds: @anatoly-scherbakov @BigBlueHat  ok with you that we make this work and merge this first without the submodule stuff?

mielvds:
> @mielvds:
> 
>     * What happened with the merge of [Update aiohttp support for Python 3.14. #208](https://github.com/digitalbazaar/pyld/pull/208)?  Confusing why it's listing all the commits here twice now on top of the aiohttp commit.
> 

oh jeez, that's not right. My bad!  I guess I wasn't sure what the strategy was. There were a couple of force-pushes to the branch and that complicated the rebasing, so I went for a merge in the end. I think I might have ended up merging both branches.

>     * Digital Bazaar has been a rebase merge organization vs having merge commits.  That is of course a topic of debate, but that's what we've been doing.  I'm guessing that's related to the above somehow.
> 

Yeah that's what I prefer too, I just got confused by git and github. I'll be more careful. Sorry!

>     * Should I do some git surgery here to fix up whatever happened?

Please do. I can try reverting the branch, but I think that better leave this to you as you know what's supposed to happen

> 
> 
> > One question about the python versions and versioning though: as it stands, we're doing 2.0.5 here. Are the breaking changes absolutely necessary here? I have a slight preference for doing some patch/minor releases first that setup testing (2.0.5) and reach spec compliance (2.1.0). I based this off on the milestones that were already there: https://github.com/digitalbazaar/pyld/milestones
> > But bundling everything in a new major release is also fine (just more complex to manage the issues).
> 
> It's been so long since a release that I think I once had trouble making testing even run on the old versions that were previously supported. I had figured we'd drop saying old versions were supported, mark it a major version to avoid any potential issues there, and move on with other changes. I look at that from the js semver policy though. Not sure if python community does it different.

Yeah makes sense. 3.0.0 is probably best; I suggest I rename the 2.0.5 milestone to v3.0.0b-testing (or v3.0.0 or something like that) and merge the remaining two in v3.0.0 (or v3.1.0)

> 
> I'm not going to look into if those aiohttp changes work with old versions, and certainly not bother to fix compat issues. Maybe that alone is enough of a breaking change?
> 
> Pardon if I missed an overall plan. I had kind of thought a major version with the workflow testing and that dropped old python support would be the next release. Other stuff could be done afterwards. But maybe some other path is better.

No it's perfectly fine, I just tried to make sense of what was already there. It's not that different either, v2.0.5 now becomes v3.0.0. I just wonder if it make sense to have a new major release with failing spec tests. A 3.0.0 with spec compliance would communicate so much better.

PR #186, opened 2023-11-12
Handle keys dropped from JSON-LD document during expansion (closes #50)

### LINKED ISSUE #50: It is not always relevant to silently ignore unrecognized properties

anatoly-scherbakov:
- Add submodules
- Update `.gitignore`
- Use spec tests from git modules as default test targets
- `list[…]` → `List[…]` because this form is for Python ⩾ 3.10 only
- Note about Git submodules in `README.rst`
- Document addition of `on_key_dropped` argument and its behavior
- Implement `on_key_dropped` argument
- Adapt tests by @pchampin to cover dropped keys handler

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,9 @@ lib/PyLD.egg-info
 profiler
 tests/test_caching.py
 tests/data/test_caching.json
+
+# Local Python version with `pyenv`
+.python-version
+
+# PyCharm & other JetBrains IDEs
+.idea
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,11 @@
+[submodule "specifications/json-ld-api"]
+	path = specifications/json-ld-api
+	url = git@github.com:w3c/json-ld-api.git
+
+[submodule "specifications/json-ld-framing"]
+	path = specifications/json-ld-framing
+	url = git@github.com:w3c/json-ld-framing.git
+
+[submodule "specifications/rdf-canon"]
+	path = specifications/rdf-canon
+	url = git@github.com:w3c/rdf-canon.git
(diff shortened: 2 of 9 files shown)


BigBlueHat:
@anatoly-scherbakov can you rebase this on the `main` branch, so it doesn't include the #182 code?

anatoly-scherbakov:
@BigBlueHat that's the reason why it's in Draft :(

It should be rebased upon #182 but Github won't allow me to do that; I think #182 has to be merged first. I will do my subsequent PRs in a proper fork of the repo. Currently, as it seems, I am using some sort of hidden fork that Github automatically creates when you try to direct your PR to upstream repo you do not control, I am not sure what exactly is going on with it.

mielvds:
@anatoly-scherbakov fyi, I included a stripped down version of your implementation here to #228

Issue #198, opened 2025-04-24
Expand() doesn't replace nested @id's with term definitions


aemnathanclinton:
We're using the CEDS Ontology, which uses opaque IRIs.  To make it easier for users to leverage SHACL and JSON-LD with this ontology, we have created a centralized context file that translates all of the opaque IRIs to human-readable IRIs.  This looks great for everything, except for the Named Individuals in the ontology act as the official list for sh:in constraints.  Here is an example of the SHACL:

```
ceds:P600546Shape
  a sh:PropertyShape ;
  sh:path ceds:P600546 ;
  sh:name "Has Person Identification Type" ;
  sh:nodeKind sh:IRI ;
  sh:in (
    ceds:NI002114100000    # PersonIdentifier - Person Identifier
    ceds:NI002114100001    # ChildIdentifier - Child Identifier
    ceds:NI002114100002    # StudentIdentifier - Student Identifier
    ceds:NI002114100003    # StaffMemberIdentifier - Staff Member Identifier
  ) ;
.
```

We have provided context aliases like so:
``` 
        "PersonIdentificationType_PersonIdentifier": {
            "@id": "NI002114100000",
            "@type": "@id"
        },
        "PersonIdentificationType_ChildIdentifier": {
            "@id": "NI002114100001",
            "@type": "@id"
        },
        "PersonIdentificationType_StudentIdentifier": {
            "@id": "NI002114100002",
            "@type": "@id"
        },
        "PersonIdentificationType_StaffMemberIdentifier": {
            "@id": "NI002114100003",
            "@type": "@id"
        }
```

We want this JSON-LD below...
"hasPersonIdentificationType": { "@id": "PersonIdentificationType_PersonIdentifier" }

...to expand into "http://ceds.ed.gov/terms#NI002114100000"

The expand code is currently ignoring the alias and applying the @base without the translation.  I was able to get it working by removing a IF condition from expand, but I'm worried about what this would break.  Here's the before & after in jsonld.py, line 5472:

before:

`     if vocab and value in active_ctx['mappings']:`

after:

`     if value in active_ctx['mappings']:    `

The problem is that vocab in this pass through the method is defaulting to false. 

After making this change, everything worked as expected.  Am I doing something wrong to make this aliasing work, or is this a valid change?  I'm happy to make a pull request if it's good. 

Thanks!

aemnathanclinton:
[...]

PR #199, opened 2025-05-05
Enable full Term Definition translation for @id values

aemnathanclinton:
Resolves issue #198


mielvds:
Closing because the original issue #198 was because of user error

PR #201, opened 2025-09-30
Fixed bug where @vocab terms with identical context conflict in context resolution

tsdrone:
Threaded a active_property "path" array through compaction, expansion and context processing such that it is added to the cache-key (normalized context representation) for storing and retrieving processed contexts from the cache.

To reproduce the bug compact with a context containing  2 property terms  of `@type`: `@vocab` with identical property-scoped '@context' declarations.


mielvds:
This PR is superseded by #249, which implements a simpler, less-intrusive fix

PR #207, opened 2025-11-04
Separate and test URI resolving 

mielvds:
This PR moves functionality from `context_resolver.py` to `iri_resolver.py` so it can be maintained and tested separately.  
This includes:
- isolating the resolving of relative to absolute IRIs in a separate `resolve()` function and adding unittests
- moving the existing function `remove_dot_segments()` and adding unittests

This work is inspired by https://github.com/rubensworks/relative-to-absolute-iri.js (which is used in https://github.com/rubensworks/jsonld-streaming-parser.js) and can be considered a JS to Python port. The unittests are also ported from that repo as they battle-test both functions. 

This PR is work in progress and currently uses pytest for convenience (this might change later).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,20 @@
   - Minimize async related changes to library code in this release.
   - In sync environment use `asyncio.run`.
   - In async environment use background thread.
-- The default test manifests or directories now default to the `./specifications` directory.
-- Add ability to run test suites using pytest and make pytest the default way for running (unit)tests.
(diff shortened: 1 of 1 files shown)


davidlehn:
> * In jsonld.js I think there was a similar move to a `url.js` file.  Someday I want to see if the js `URL` api can replace a lot of this.  Not sure if there's a similar library for python or what considerations there are using such a thing.

I now see the next related PR has a move to a more standard library.

Issue #181, opened 2023-10-18
Include JSON-LD specifications as `git` submodules


anatoly-scherbakov:

BigBlueHat:
My one concern here is that git submodules need to be updated if/when the upstream repo changes as they're pinned to specific commit SHAs. It may be all that's needed is a simple python script to run the commit commands needed to pull the latest repos locally.

anatoly-scherbakov:
@BigBlueHat Indeed, periodically one has to update the submodules. I would argue this is a good thing, similar to how one has to explicitly update Python dependencies of a project: it makes the project more reliable, and the builds more stable.

Updating dependencies, including the specs, should be a conscious decision and a human should be in the loop to fix possible issues.

Otherwise, a test run for a completely unrelated change can throw a bunch of errors clearly unrelated to the change, and it will take time to realize that one of the specs (and which one?) has been meanwhile updated.

Issue #191, opened 2024-02-03
Use `Makefile` to improve developer experience


anatoly-scherbakov:
https://github.com/digitalbazaar/pyld/pull/182 proposes to use a `Makefile` to automate the project. This issue provides reasoning for introducing that tool.

## Why is that important?

An open source library has software developers as its target audience:

* as users,
* and as contributors.

To attract contributors, it likely benefits to improve their developer experience -- to ensure they enjoy the process of contributing to the project.

One of aspects of that could be saving them from tedious repetitive tasks, and Make can do that.

## Updating submodules versions

```
git submodule update --remote --init --recursive
```

is rather lengthy and tedious to copy-paste from `README.md` all the time, -- and thus provides a better developer experience.

## Running tests

We could do

```
make test
```

instead of, say, `python tests/runtests.py`.

## CI

Such commands as `make test` can also be reused in GitHub Actions CI (assuming we do want to use those), thus reducing the repetition in the overall project repo.

PR #209, opened 2025-11-09
Add git submodules for specs

anatoly-scherbakov:
I think my previous PR was trying to change too much and I decided it would be easier to recreate the most important changes in a new one.

This PR:

* Creates two submodules (as discussed, `normalization` which is now `rdf-canon` is out of scope for now),
* Wires them into CI (thanks @davidlehn for introducing GitHub Actions!)

- **#181 Add submodules for JSON-LD API & Framing specifications**
- **#181 Use spec git submodules in CI**

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  # Update specifications' submodules
+  # https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference
+  - package-ecosystem: "submodules"
+    directory: "/"
+    schedule:
+      interval: "weekly"


mielvds:
It kind of changes the test system, because contributors now need to init submodules instead of cloning them (although they still could). I'm all for small PR's, but I also like them to include documentation about the changes they introduce. It's a balance.
Also based on @davidlehn's concern in the previous PR: 

> Adding this feature forces the maintainers to learn to use it. Help us (me) out with a few more snippets of docs perhaps in the README and/or CONTRIBUTING. Maybe even more script/makefile targets to run. Specifically, what do we (I) need to type to keep the test suites synced up all the time.

So makefile is covered by #191, but I would add the snippets here because it's not that much work (I can add this).

anatoly-scherbakov:
> need to init submodules instead of cloning them

After this PR, contributors are not yet _required_ to do so, they might continue cloning specs separately as before. This PR prepares the repo for a submodule-aware workflow.

In a follow up PR, I will:

* Update `runtests.py` to use submodules by default;
* Update documentation,

if that makes sense. Let me know what you think.

PR #218, opened 2025-12-05
Explain submodules in README and bring back normalization

mielvds:
This PR 
- explains the use of submodules in the README.rst 
- brings back the removed normalization test-suite for the time being to avoid big changes.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -190,14 +190,18 @@ Tests
 This library includes a sample testing utility which may be used to verify
 that changes to the processor maintain the correct output.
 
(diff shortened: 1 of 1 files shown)


davidlehn (approved):
Approving to move things forward.  Thanks for the docs.

The "sibling" terminology needs to change.  I used that because those dirs were actually siblings to the pyld dir.  And that made sense with the setup I have of a top level "json-ld" dir with specs, implementations, and other files in one place.  Pointing to a sub-sub-dir is something else.

PR #222, opened 2025-12-12
Simplify compare functions for sorting

mielvds:
This PR removes some of 'old-style' python code and fixes linter issues:
- remove the need for the `cmp()` function wrt to sorting
- use `isinstance()` and `is` to do type checking (instead of  for instance `type(x) == bool`)
- change `not X in Y` to `X not in Y`
- always raise specific error

### DIFF
--- a/lib/pyld/jsonld.py
+++ b/lib/pyld/jsonld.py
@@ -4346,7 +4346,7 @@ class JsonLdProcessor(object):
                         idx = options['link'][id_].index(input_)
                         # already visited
                         return options['link'][id_][idx]
-                    except:
+                    except ValueError:
                         # prevent circular visitation
                         options['link'][id_].append(input_)
                 else:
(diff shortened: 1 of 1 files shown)


lib/pyld/jsonld.py
```diff
@@ -6060,19 +6049,7 @@ def permutations(elements):
...
-
-    :param a: the first string.
-    :param b: the second string.
 
-    :return: -1 if a < b, 1 if a > b, 0 if a == b.
-    """
-    rval = cmp(len(a), len(b))
-    if rval == 0:
-        rval = cmp(a, b)
-    return rval
```
davidlehn: A few more blank lines than needed here.

davidlehn (commented):
- Thanks for cleanup.
- I think much this style was originally due to manual js to py conversion.

PR #223, opened 2025-12-16
Isolate parsing/serialising NQuads and canonicalisation algorithms in indivual modules

mielvds:
This PR moves the canonicalization algorithms, the nquads parsing and the `IdentifierIssuer` in a separate modules. This is the first step of eventually moving the canonicalization algorithms out of this library.

Because it no longer makes sense that an nquad parser or canonicalization algorithm raises JSON-LD errors, they have their own neutral `UnknownFormatError` and `ParserError`.

### DIFF
--- a/lib/pyld/canon.py
+++ b/lib/pyld/canon.py
@@ -1,6 +1,6 @@
 
 import hashlib
-from pyld.nquads import parse_nquads, to_nquad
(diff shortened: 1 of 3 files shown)


mielvds:
You're right @davidlehn, I split off the exception chaining in #224. The nquads API doesn't change because I kept the methods. For the canon and IdentifierIssuer, I added entries to the changelog

PR #227, opened 2025-12-21
Try out `ruff`

anatoly-scherbakov:
# Why

We do not have a code formatter that would help us unify the codebase.

# What

Implement `ruff`. Run it in CI, but only against one file: to taste it and to reduce the probability of git conflicts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,10 +9,7 @@ test:
 upgrade-submodules:
 	git submodule update --remote --init --recursive
 
-# At this stage, we are limiting our formatting efforts to one file. We need to ensure that:
-# * our formatting rules are sane,
-# * we are not introducing conflicts with currently open PRs,
-# * and the PR introducing `ruff` is not too large.
+# TODO: Expand to lib/ and tests/ as linting issues are resolved.
 RUFF_TARGET = lib/pyld/context_resolver.py
 
 lint:


Makefile
```diff
@@ -7,4 +7,17 @@ test:
...
 upgrade-submodules:
-	git submodule update --remote --init --recursive
\ No newline at end of file
+	git submodule update --remote --init --recursive
+
+# At this stage, we are limiting our formatting efforts to one file. We need to ensure that:
+# * our formatting rules are sane,
+# * we are not introducing conflicts with currently open PRs,
+# * and the PR introducing `ruff` is not too large.
+RUFF_TARGET = lib/pyld/context_resolver.py
```
davidlehn: - This contemporary commentary isn't needed.  Looks more like PRs comments than code ones.
- Might as well point it at everything so the results (or mess!) can be seen.
- I assume it can't be too bad, at least for checks?  Other than perhaps line length.  If it is, maybe tune/comment the config temporarily until those changes can be addressed?

mielvds: I agree with these points. For the last point though: could we turn this around and see how involved the changes actually are? It might be unnecessary to do this incrementally, which will save us some time.

pyproject.toml
```diff
@@ -0,0 +1,57 @@
+[tool.ruff]
+# Target Python version
+target-version = "py310"
+
+# Line length (matching rdflib's flake8 config)
+line-length = 88
```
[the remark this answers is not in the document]
anatoly-scherbakov: Change of this to 79 will reduce total number of linter errors from 116 to 114. I do not believe this is a big difference, is it?

Makefile
anatoly-scherbakov: If directories which `make lint` and `make fmt` point to are different, then:

  * Either CI fails on errors which `make fmt` is unable to fix,
  * Or `make fmt` changes code on which `make lint` does not complain.

Both of which are counter-intuitive outcomes for developers trying to contribute to this project.

I believe the comment I added failed to convey the purpose, for which I apologize. `ruff` throws **116 linting errors** against `lib` directory, which is too much for one PR, in my opinion. In addition, there are large PRs in development as far as I recall. If I were to reformat the whole codebase in this PR, there would be a lot of merge conflicts introduced. Basically, it would have rendered this PR unmergeable, for the time being.

In its current form, the PR should be mergeable because it only touches one piece of the code and changes in that piece are very minimal.

When and if this one is merged, I will make a next PR which will expand the scope of the checks.

In addition, I didn't want the automatically reformatted code to obscure the changes in configs and CI - things which bear some meaning and need a review. A mix up of lots of automatic changes plus a bit of manual changes is an unpleasant dish for a reviewer.

Issue #187, opened 2023-11-17
IRI expansion with missing `@base` does not conform to RFC 3986


RinkeHoekstra:
[RFC 3986 section 5.1](https://datatracker.ietf.org/doc/html/rfc3986#section-5.1) specifies that relative URIs should be expanded against the document's base URI. In absence of an explicit base, there are prescribed steps to determine the base IRI for a given document:

[5.1.1](https://datatracker.ietf.org/doc/html/rfc3986#section-5.1.1).  Base URI Embedded in Content . . . . . . . . . . [29](https://datatracker.ietf.org/doc/html/rfc3986#page-29)
[5.1.2](https://datatracker.ietf.org/doc/html/rfc3986#section-5.1.2).  Base URI from the Encapsulating Entity . . . . . [29](https://datatracker.ietf.org/doc/html/rfc3986#page-29)
[5.1.3](https://datatracker.ietf.org/doc/html/rfc3986#section-5.1.3).  Base URI from the Retrieval URI  . . . . . . . . [30](https://datatracker.ietf.org/doc/html/rfc3986#page-30)
[5.1.4](https://datatracker.ietf.org/doc/html/rfc3986#section-5.1.4).  Default Base URI . . . . . . . . . . . . . . . . [30](https://datatracker.ietf.org/doc/html/rfc3986#page-30)

The current implementation in pyLD ignores the last two requirements. For 5.1.3 this is understandable, as the library only operates on a data payload. However, 5.1.4 is the catch-all that would ensure that `@id` values are always expanded to absolute IRIs. 

In absence of this, non-IRI `@id` values in documents that do not explicitly specify a base in a context are not expanded to an absolute IRI. This means that the `to_rdf` function ignores them when producing N-Quads output. This is a showstopper for <https://github.com/RDFLib/rdflib/issues/2308>.

The JSON-LD spec does allow for a means to prevent expansion against a base by setting `@base` to `null` (see <https://www.w3.org/TR/json-ld/#base-iri>) but does not specify that `null` is the default.

This violates test [t0060 in](https://w3c.github.io/json-ld-api/tests/expand/0060-in.jsonld) and [t0060](https://w3c.github.io/json-ld-api/tests/expand/0060-out.jsonld).

The output should be something similar to (with a different application-specific base):

```
[
  {
    "@id": "https://w3c.github.io/json-ld-api/tests/document-relative",
[...]

Issue #149, opened 2021-02-18
Nested "@id" elements are not processed correctly in framing


hsolbrig:
Changing the `@id` alias in an inner context doesn't work correctly.

Synopsis:
```json
{   "@context": {
            "namespace": "@id", 
             "contents": {
                  "@context": {
                       "uri": "@id
                  }
              }
      }
}
```

Doesn't work correctly in framing -- both the inner and the outer element are returned as `"namespace"`

The attached file is a complete unit test.

[test_ld_11_issue.txt](https://github.com/digitalbazaar/pyld/files/6006329/test_ld_11_issue.txt)

cmungall:
Would be great to see this fixed!

mielvds:
This seems to fixed according to the test in #228

PR #228, opened 2025-12-23
Merge tests and fixes from stale PRs

mielvds:
This PR attempts to integrate tests and fix from other PRs. Some of them are very old and are not mergeable anymore. However, most of these issues still persist and to not let this work go to waste, they are ported here:

Includes tests from or for issues:

- [x] https://github.com/digitalbazaar/pyld/issues/50
- [x] https://github.com/digitalbazaar/pyld/issues/187
- [x] https://github.com/digitalbazaar/pyld/issues/149
- [x] https://github.com/digitalbazaar/pyld/pull/31
- [x] https://github.com/digitalbazaar/pyld/pull/59
- [x] https://github.com/digitalbazaar/pyld/pull/202

Includes fix from or for:

- [x] https://github.com/digitalbazaar/pyld/pull/51
- [x] https://github.com/digitalbazaar/pyld/issues/187
- [x] https://github.com/digitalbazaar/pyld/issues/149
- [x] https://github.com/digitalbazaar/pyld/pull/31
- [x] ~~https://github.com/digitalbazaar/pyld/pull/60~~ --> introduces regressions, so no included
- [x] https://github.com/digitalbazaar/pyld/pull/202

### DIFF
--- a/tests/runtests.py
+++ b/tests/runtests.py
@@ -859,10 +859,7 @@ class EarlReport():
 # supported test types
 TEST_TYPES = {
     'jld:CompactTest': {
-        'pending': {
-            'idRegex': [
-            ]
-        },
+        'pending': {},
         'skip': {
             # skip tests where behavior changed for a 1.1 processor
             # see JSON-LD 1.0 Errata
@@ -883,10 +880,7 @@ TEST_TYPES = {
         ]
     },
     'jld:ExpandTest': {
-        'pending': {
-            'idRegex': [
-            ]
-        },
+        'pending': {},
         'runLocal': [
             '.*remote-doc-manifest#t0003$',
             '.*remote-doc-manifest#t0004$',
@@ -924,10 +918,7 @@ TEST_TYPES = {
         ]
     },
     'jld:FlattenTest': {
-        'pending': {
-            'idRegex': [
-            ]
-        },
+        'pending': {},
         'skip': {
             # skip tests where behavior changed for a 1.1 processor
             # see JSON-LD 1.0 Errata
@@ -945,10 +936,7 @@ TEST_TYPES = {
         ]
     },
     'jld:FrameTest': {
-        'pending': {
-            'idRegex': [
-            ]
-        },
+        'pending': {},
         'skip': {
             # skip tests where behavior changed for a 1.1 processor
             # see JSON-LD 1.0 Errata
@@ -1005,7 +993,6 @@ TEST_TYPES = {
                 '.*toRdf-manifest#ttn02$',
                 # well formed
                 '.*toRdf-manifest#twf05$',
-                '.*toRdf-manifest#twf06$',
                 # uncategorized
                 '.*toRdf-manifest#tc038$',
                 '.*toRdf-manifest#ter54$',


tests/test_jsonld.py
```diff
@@ -0,0 +1,498 @@
...
+        }
+
+        result = jsonld.to_rdf(input)
+        assert result == expected
+
+
+class TestCompact:
+    # Issue 59 - PR: https://github.com/digitalbazaar/pyld/pull/60
+
+    def test_simple_compaction(self):
```
davidlehn: Is this test covered by the main test suite?  Could probably be moved there if not.

davidlehn:
General:
- I'm not sure this was the best approach to handle old issues and PRs.  It's now very difficult to review, and difficult to handle one piece at a time.
- I don't think old issues should be closed until the issue is actually fixed, or at least merged into main.

Dropped "keys" / logging:
- The issue addressed by the `on_key_dropped` callback patch likely needs to be more involved.  That was the path jsonld.js had to start.  It then moved over to "safe mode" to handle that and many other situations that could cause problems when signing canonized output.
- If I had gotten to implementing a pyld "safe mode", it would likely have taken the simple path of throwing errors aligned with the js ones in all the needed places.  More flexibility may only be needed for developer tools, which could be future work.
- Not sure what to suggest should be done here.  This patch only handles one situation of many needed for a full "safe mode".  An important one though.
- Is debug logging is the best default?  Why not a no-op?  The spec does allow dropped properties so arguably that's not something that needs debugging.  If dropped properties are always an error is up for debate.
- On naming, I think "property" might be better than "key"?  It aligns more with the spec.  For the situation here, I think the js code is using a "invalid property" event / error code.  But I understand the desire to match API naming with native data structures.

mielvds:
Hi @davidlehn 

> General:
> 
> * I'm not sure this was the best approach to handle old issues and PRs.  It's now very difficult to review, and difficult to handle one piece at a time.
> * I don't think old issues should be closed until the issue is actually fixed, or at least merged into main.

I agree that this is not ideal and I admit I'm being pragmatic here. But this repo has too many open PR's and issues ranging from 2 to 9 years that could no longer be merged. So this is mostly housekeeping. But noted: I'll keep the issues open until they are actually fixed and merged.

I selected and checked this set of issues carefully because they are about valid issues, include tests and partial fixes. All spec tests still pass like they did before. All in all, I think the changes are not too big and reviewing should still be doable. Best is to start with checking the tests. I will add some more explanatory comments to help understand the original issue/fix.

> 
> Dropped "keys" / logging:
> 
> * The issue addressed by the `on_key_dropped` callback patch likely needs to be more involved.  That was the path jsonld.js had to start.  It then moved over to "safe mode" to handle that and many other situations that could cause problems when signing canonized output.

I agree but could we handle this is a new issue? I suggest we keep this change and possibly shake things up in a new release. I'm interested in how jsonld.js handles this.


> * If I had gotten to implementing a pyld "safe mode", it would likely have taken the simple path of throwing errors aligned with the js ones in all the needed places.  More flexibility may only be needed for developer tools, which could be future work.

This is a good approach. Can we place handlers (in line with `on_key_dropped`) where needed and implement a safe mode on top of that? This would also give developers more fine-grained control. 

> * Not sure what to suggest should be done here.  This patch only handles one situation of many needed for a full "safe mode".  An important one though.

I suggest a new issue to discuss a new iteration. I can open this and summarize what has been said so far.

> * Is debug logging is the best default?  Why not a no-op?  The spec does allow dropped properties so arguably that's not something that needs debugging.  If dropped properties are always an error is up for debate.

I actually agree, I just wanted to stay true to the original implementation. Introducing a logger is probably also something for a different PR. I'll patch this.

> * On naming, I think "property" might be better than "key"?  It aligns more with the spec.  For the situation here, I think the js code is using a "invalid property" event / error code.  But I understand the desire to match API naming with native data structures.

Hmm good point; I'll make the change. An explanation about 'key' in the docs should suffice

mielvds:
@davidlehn I added the changes you proposed in your last two points. I also added a docstring to every test and added comments explaining all changes. Would this suffice for a review and merge?

Issue #220, opened 2025-12-09
Moving out RDF Dataset Canonicalization


mielvds:
PyLD contains implementations for a couple of RDF dataset canonicalization algorithms from back when this was part of the JSON-LD work, as pointed out by @davidlehn in https://github.com/digitalbazaar/pyld/issues/190. Since then, this standardization work has moved to the Credentials Community Group, which then evolved into the rdf-canon Working Group. Eventually, this led to the RDF Dataset Canonicalization w3c recommendation. 

All of this produced three algorithms:
- [URDNA2012]() implemented at https://github.com/digitalbazaar/pyld/blob/58ea0ea9e64c87b5ab8da78016b1cae170700ddf/lib/pyld/jsonld.py#L6073
- [URDNA2015](https://www.w3.org/community/reports/credentials/CG-FINAL-rdf-dataset-canonicalization-20221009/) implmented at https://github.com/digitalbazaar/pyld/blob/58ea0ea9e64c87b5ab8da78016b1cae170700ddf/lib/pyld/jsonld.py#L5649
- [RDFC-1.0](https://www.w3.org/TR/rdf-canon/#dfn-rdfc-1-0) which is still to be implemented, but only [deviates slightly from `URDNA2015`](https://www.w3.org/TR/rdf-canon/#urdna2015).

There have been some discussions/efforts (https://github.com/digitalbazaar/pyld/pull/182,https://github.com/digitalbazaar/pyld/pull/209,https://github.com/digitalbazaar/pyld/issues/190) on whether this should move elsewhere. RDF Dataset Canonicalization has a broader use and could therefore benefit from being a repository of its own; the PyLD code would be more focused. Digitalbazaar actually already did this for its JS counterpart.  This would also highly benefit RDFLib, which does not have a RDFC-1.0 implementation, but already provides methods to compare graphs. 

In light of this library eventually moving to the RDFlib ecosystem, I propose that maybe this functionality moves first? 
A possible path forward would be:
1. Preparing a PR isolating the implementation and tests in a separate module (I can do this)
2. Adding the RDFC-1.0 implementation and test-suite (I can try this, but maybe with help of the broader RDFLib community)
3. Copying this part of the code to a RDFLib based repo and publish the package on pypi
4. Replacing the code with the new dependency

(CC @davidlehn @BigBlueHat @nicholascar @edmondchuc @anatoly-scherbakov )

BigBlueHat:
[...]

PR #245, opened 2026-02-25
Implement RDFC-1.0

mielvds:
This PR is a rather big one. It fully implements URDNA2015, URDNA2012 and RDFC10 and has complete test-suite coverage on these algorithms (with the exception of the one on poisonous datasets, but more about that later). It 'fixes' #220 

Some general remarks before I dive into the details:
- I went back and forth on how approach this: move the canon stuff to a new repo first, or fix the problems first. I sided with the latter as the test harness is already in place and the code is too tangled to not break everything.
- This **introduces rdflib into PyLD**, but limited to the normalization/canonicalization `canon.py`. This made the code much much cleaner, but ironically didn't fix what I introduced it for: nquad serialization. The relevant methods are copied over from rdflib until I can turn them into PRs for rdflib. 
- I added some functions to transform the legacy RDF.JS dataset structure in an `rdflib.Dataset` and back. This should ensure backwards-compatibility on some methods such as `jsonld.normalization()` and `URDNA2015.main()`.
- While switching to rdflib introduces significant changes, I tried to **change as little as possible otherwise**. Optimizing the algorithm implementations can be done later.

Overview of the changes:
- added `rdflib` dependency in all relevant config and code files.
- the change to `rdflib` mostly changed
  - RDF term type checking (e.g., checking is something is a bnode)
  - looping over triples/quads
  - serialization
  - constructing RDF terms (custom deepcopy is no longer needed)
- `util.py`
  - added the functions `from_legacy_dataset()` and `to_legacy_dataset()` to easily go from an `rdflib.Dataset` to an RDFJS-like `dict` wherever needed. 
  - Also added unittests in `tests/test_util.py` for these functions. We might want to move more general-purpose methods to this file.
- `canon.py`:
  - the main logic was moved to `URDNA2015._canonicalize(self, dataset: Dataset)` while handling input and output remained in `URDNA2015.main()`. The latter now does the nquads parsing instead of `JsonLdProcessor.normalize()`, so all parsing and serialization is handled by the same class. 
  - the method `URDNA2015._canonicalize(self, dataset: Dataset)` accepts an `rdflib.Dataset` and returns a tuple with 
    - the canonicalized result as a nquads `str` and 
    - the blank node identifier map as `dict`.
  - The method `URDNA2015 .main(self, dataset: str | dict | Dataset, options)` now accepts a `rdflib.Dataset` object in addition to an nquads `str` or the original RDFJS-like`dict`. It returns 
    - a `str`: the serialized nquads result, or 
    - a `dict`: the result as RDFJS-like dataset or the blank node identifier map when the new parameter `outputMap` is `True`.
  - the hashing algorithm is now an class attribute `URDNA2015.hash_algorithm` so it easily configurable (required for RDFC1.0)
  - the `permutations()` function now uses `itertools.permutations` instead of a custom implementation.
  - replacements for rdflib's `_nq_row` and `_quoteLiteral`, which should eventually move to a fix for rdflib's nquads serializer.
- `tests/runtests.py`
  - (re-)enabled all skipped URDNA2015, URDNA2012
  -  Added the RDFC10 tests
    - added support for testing blank-node identifier maps. If the result of a test is a dict and the expected value is a string, the expected value is now parsed as JSON.
    - added support for testing with different hashing algorithms
    - !! I did not include test 74c on dataset poisoning because I'm not sure how to handle that yet. We need to discuss possible guardrails  first before continuing on this.

It's possible that we move this code out before ever merging, but at least it can be easily tested. That said, when importing this functionality as an external library, we include rdflib anyway. It therefore makes sense to continue replacing the RDFJS-like data structures elsewhere in the code, essentially making it fully rdflib compliant and dependent.

Since this involves RDFLib: @nicholascar, anyone who can maybe have a look at this? Also @davidlehn @BigBlueHat @anatoly-scherbakov

### DIFF
--- a/lib/pyld/canon.py
+++ b/lib/pyld/canon.py
@@ -508,6 +508,14 @@ class URGNA2012(URDNA2015):
     def create_hash(self):
         return hashlib.sha1()
 
+class RDFC10(URDNA2015):
+    """
+    RDFC10 implements the RDF Canonicalization algorithm version 1.0.
+    """
+    # TODO: Stub that uses URDNA2015 for now
+    def __init__(self):
+        URDNA2015.__init__(self)
+
 
 def permutations(elements):
     """
--- a/lib/pyld/canon.py
+++ b/lib/pyld/canon.py
@@ -2,7 +2,7 @@ import copy
 import hashlib
 
 import rdflib
-from rdflib import XSD, BNode, Dataset, Literal, Node, URIRef
+from rdflib import XSD, BNode, Dataset, Literal, Node
 from rdflib.graph import DATASET_DEFAULT_GRAPH_ID
 from rdflib.plugins.serializers.nt import _quote_encode
 
@@ -29,7 +29,7 @@ class URDNA2015:
             and options['format'] != 'application/nquads'
         ):
             raise UnknownFormatError('Unknown output format.', options['format'])
-        
+
         # handle differtent input types: nquads string, dict (legacy ), or rdflib Dataset
         rdflib_dataset = Dataset()
         if isinstance(dataset, str):
@@ -47,21 +47,21 @@ class URDNA2015:
             rdflib_dataset = dataset
         else:
             raise ValueError(f'Unsupported dataset type: {type(dataset)}')
-        
+
         normalized = self._main(rdflib_dataset)
         with open(options.get('algorithm') + '.nq', 'w') as f:
             print(normalized, file=f)
-    
+
         # 8) Return the normalized dataset.
         if (
             options.get('format') == 'application/n-quads'
             or options.get('format') == 'application/nquads'
         ):
             return normalized
-        
+
         result = Dataset().parse(data=normalized, format='nquads')
         return to_legacy_dataset(result)
-        
+
     def _main(self, dataset: Dataset) -> str:
         self.dataset = dataset
         # 1) Create the normalization state.
@@ -196,7 +196,7 @@ class URDNA2015:
             p_n = p # Predicates are never BNodes in standard RDF
             o_n = map_node(o)
             g_n = map_node(g)
-            
+
             # Use modified version of rdflib's internal _nq_row for standardized string output
             line = self._nq_row((s_n, p_n, o_n),g_n)
 
@@ -240,7 +240,7 @@ class URDNA2015:
             s_n = self.modify_first_degree_component(id_, s)
             o_n = self.modify_first_degree_component(id_, o)
             g_n = self.modify_first_degree_component(id_, g)
-            
+
             # Use rdflib's internal _nt_row for standardized string output
             line = self._nq_row((s_n, p_n, o_n), g_n)
             nquads.append(line)
@@ -584,12 +584,32 @@ class RDFC10(URDNA2015):
             return f"{encoded}"
 
     def _quote_encode(self, l_: str) -> str:
-        return '"{}"'.format(
-            l_.replace("\\", "\\\\")
-            .replace("\n", "\\n")
-            .replace('"', '\\"')
-            .replace("\r", "\\r")
-        )
+        # Accept either an rdflib Literal or a plain string
+        s = str(l_)
+
+        parts = []
+        for ch in s:
+            code = ord(ch)
+            if ch == "\\":
+                parts.append('\\\\')
+            elif ch == '"':
+                parts.append('\\"')
+            elif ch == "\n":
+                parts.append('\\n')
+            elif ch == "\r":
+                parts.append('\\r')
+            elif ch == "\t":
+                parts.append('\\t')
+            elif code == 0x08:  # backspace
+                parts.append('\\b')
+            elif code == 0x0C:  # form feed
+                parts.append('\\f')
+            elif code == 0x0B or (code < 0x20) or (code == 0x7F):  # vertical tab -> use \u000B
+                parts.append(f'\\u{code:04X}')
+            else:
+                parts.append(ch)
+
+        return '"' + ''.join(parts) + '"'
 
 
 def permutations(elements):
--- a/lib/pyld/canon.py
+++ b/lib/pyld/canon.py
@@ -7,6 +7,8 @@ from rdflib.graph import DATASET_DEFAULT_GRAPH_ID
 from rdflib.plugins.serializers.nt import _quote_encode
 
 from pyld.identifier_issuer import IdentifierIssuer
+from pyld.util import from_legacy_dataset, to_legacy_dataset
+
 
 # TODO: use drop-in replacement to not serialize with xsd:string
 def _nq_row(triple, context):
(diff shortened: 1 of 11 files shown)


lib/pyld/canon.py
```diff
@@ -14,44 +20,82 @@ def __init__(self):
...
+        elif isinstance(dataset, Dataset):
+            rdflib_dataset = dataset
+        else:
+            raise ValueError(f'Unsupported dataset type: {type(dataset)}')
+
+        normalized, bnode_map = self._canonicalize(rdflib_dataset)
+
+        # Merge any new bnode IDs from the parser into the id_map,
+        # mapping old bnode IDs to their new canonical IDs
+        if parser:
```
WhiteGobo: Doesnt this result in a nameerror because parser is only set when isinstance(dataset, str)?

lib/pyld/canon.py
```diff
@@ -504,57 +580,73 @@ def create_hash_to_related(self, id_, issuer):
...
+                raise UnknownFormatError('Unknown hash algorithm.', hash_algorithm)
+            self.hash_algorithm = getattr(hashlib, hash_algorithm.lower())
+
+    def _quoteLiteral(self, l_: Literal) -> str:  # noqa: N802
+        """A simpler version of term.Literal.n3()"""
+
+        encoded = self._quote_encode(l_)
+
+        if l_.language:
+            if l_.datatype:
```
WhiteGobo, line 603: I would remove this test `if l_.datatype`: in rdflib, if language is given, the datatype should always be forced, so this test for datatype should always return the same. As far as i remember the correct datatype for lang tagged literals is rdf:langstring, so there might be in future a change to return rdfs:langstring instead of None.

WhiteGobo (commented):
I just looked over the parts, where the behaviour of rdflib plays a role, because i dont fully understand what the code itself does. I hope my comments are a little helpful.
I have another more generell comment, as BNode is a subclass to str if you eg. check types with mypy you might get no error, when you by mistake compare a str to a BNode. But this would always fail. this might play a role, when you call hash_related_blank_node. But as i said, i have not a full grasp on the program.

lib/pyld/canon.py
mielvds: Agreed! But this is code copied (and cleaned) from rdflib's nquads serializer, so I'll mix it in a PR there.

mielvds:
> as BNode is a subclass to str if you eg. check types with mypy you might get no error, when you by mistake compare a str to a BNode. But this would always fail. this might play a role, when you call hash_related_blank_node. But as i said, i have not a full grasp on the program.

@WhiteGobo do you mean `isinstance(component, BNode)`? what would be a better way to check if something is a blank node?

WhiteGobo:
> do you mean isinstance(component, BNode)?

No i did think more of finding type errors with tools like mypy. And i think of the over way aroung so something like `isinstance(accidently_some_BNode, str)`.

This was merely a warning, when using `str` instead of directly using a `BNode`, that mypy wouldnt find inconsistent use between those two.
Just to illustrate take for example canon.py line 55ff. I will change some lines, which would produce an error:
```python
        bnode_map: "dict[str, str]" # this would be used, so that mypy can detect wrong inserts.
        normalized, bnode_map = self._canonicalize(rdflib_dataset)
        # mapping old bnode IDs to their new canonical IDs
        for k, v in parser._bnode_ids.items():
            bnode_id = str(v)
            # Imagine, you were unconcentrated and you used by mistake the original bnode instead of the extracted str.
            if v in bnode_map: # This line is errornous and therefore mypy should throw a warning
                bnode_map[k] = bnode_map[bnode_id]
                del bnode_map[bnode_id]
```

For these kind of scenario, i use mypy to warn me, that i did something wrong, but mypy would register BNode as valid value, because, its just a subclass to `str`. And i dont mean, that extracting the id from a BNode is wrong, i just wanted to point it out, if you use mypy to find such errors. As you extract `bnode_map` from somewhere else, i dont even think, you should prefer `BNode` over `str`.

mielvds:
ah ok, type checking, gotya. Adding type hints everywhere would generate too many changes (and be too much work). I think this for some other PR

gjhiggins:
I know it's not my horse but FYI ... in the course of adapting your implementation as an RDFLib plugin, I encountered use case 9 in the w3c [documentation of the spec](https://www.w3.org/TR/2024/REC-rdf-canon-20240521/#duplicate-paths-ex) and I don't seem to be able to replicate the results using the code in this PR. If my understanding of your example is correct and the following snippet is valid, the other unexpected result is that the resulting id_map flips arbitrarily between two canonicalizations ...

```python
from pyld.canon import RDFC10
data = '_:e0 <http://purl.org/base#p1> _:e1 .\n' + \
'_:e1 <http://purl.org/base#p2> "Foo" .\n' + \
'_:e2 <http://purl.org/base#p1> _:e3 .\n' + \
'_:e3 <http://purl.org/base#p2> "Foo" .\n'

expected = {'e0':  'c14n0,', 'e1':  'c14n1', 'e2':  'c14n2', 'e3':  'c14n3'}

actual = [
    {'e1': 'c14n0', 'e0': 'c14n1', 'e3': 'c14n2', 'e2': 'c14n3'},
    {'e3': 'c14n0', 'e2': 'c14n1', 'e1': 'c14n2', 'e0': 'c14n3'},
]

canonicalized = RDFC10().main(data, dict(
    inputFormat='application/n-quads', outputMap=True)
)

assert canonicalized in actual
```

I'd be temped to ascribe it to a documentation infelicity were it not for the exhaustive execution log that accompanies the use case description --- and of course, I could just be holding it wrong.

Cheers
Graham

mielvds:
> resulting id_map flips arbitrarily between two canonicalizations

Good question. It strange that this is not part of the test-suite. Could you perhaps repeat your question there, because I'm interested in what the expected behavior is here. 

In order to pass some tests, the blank nodes are mapped according to the order of the objects. This might have something to do with rdflib's parser assigning new IDs to every blanknode, which might change that order. I added your case as a unittest and will investigate a little further.

anatoly-scherbakov (approved):
The PR adds `specifications/rdf-canon` to `.gitmodules` and wires it into default spec dirs at `tests/runtests.py`, but the PR diff does not include a gitlink for the submodule. Thus, `specifications/rdf-canon` is absent, so default pytest silently skips those tests instead of running the newly added coverage.

mielvds:
oh yes, I noticed the CI was not running the tests at all. I added the command, but the submodule is indeed not being fetched. Can you fix this @anatoly-scherbakov ? I don't have much experience with this.

anatoly-scherbakov:
@mielvds committed. In the olden days, I remembered how to do this. One has to do `git submodule add` command or something like that. Editing `.gitmodules` alone won't suffice, and I did stumble upon this more than once.

Now, I must confess, I do entrust this kind of tedium to agents.

mielvds:
Well, I don't need an agent to know you must use `git submodule add`, so I must have done something weird. Anyway, thanks for the fix!

gjhiggins:
> Good question. It strange that this is not part of the test-suite. Could you perhaps repeat your question there, because I'm interested in what the expected behavior is here.

Given that the rdf-canon test suite includes explicit RDFC10MapTest `id_map` tests, I'd assume the described behaviour is precisely what is expected. (Up until today, I hadn't paid any attention to the map tests and was unaware that my adaptation of the RDFLib W3C test suite harness was omitting to run those tests. I adjusted my adaptation to recognise and actually *run* the RDFC10MapTest tests and I'm seeing one failure: the SHA384 test: eval passes, the map doesn't).

> In order to pass some tests, the blank nodes are mapped according to the order of the objects. This might have something to do with rdflib's parser assigning new IDs to every blanknode, which might change that order. I added your case as a unittest and will investigate a little further.

I became aware of the ID reassignment issue after reading @WhiteGobo's [observation](https://github.com/RDFLib/rdflib/issues/2736#issuecomment-2006718598) in the skolemization discussion but I'm doubtful that it's the source of the issue in the duplicate path RDFC10MapTest failure. After extending `**kwargs` processing to ingest the `bnode_context` dict and making a couple of adjustments to my code, I'm now able to recover the pre-vs-post id_map and check it directly - and it looks just fine and it's the approach that I use when running the RDFC10MapTests, so from that perspective, the ID reassignment doesn't appear relevant (/me ensures hat is edible).

mielvds:
That's exactly what this implementation does as well. You can pass "hashAlgorithm": "SHA384" to `main()`

gjhiggins:
> I'm seeing one failure: the SHA384 test: eval passes, the map doesn't).

Meh, operator error - had I forgotten to copy the `hash_algorithm` kwarg-and-binding into the `serialize` call params in my newly-authored maptest runner. After remediating this blunder, all the tests in the W3C rdf-canon suite now pass.

lib/pyld/canon.py
```diff
@@ -434,6 +486,32 @@ def hash_nquads(self, nquads):
             md.update(nquad.encode('utf8'))
         return md.hexdigest()
 
+    # TODO: use drop-in replacements to not serialize with xsd:string; better to solve this at the rdflib level
```
gjhiggins, line 489: The code in this `def _nq_row` is good to go as an RDFLib change, it can directly replace the implementation in [LoC#49](https://github.com/RDFLib/rdflib/blob/0fdac395078b81313f30eeab74b1a18eb7bb4db6/rdflib/plugins/serializers/nquads.py#L49) 

However, the `def _quoteLiteral` causes test failures in the RDFLib test suite:

- unchanged: `7668 passed, 581 skipped, 329 xfailed, 36 xpassed, 36196 warnings`
- changed: `58 failed, 7610 passed, 581 skipped, 329 xfailed, 36 xpassed, 36234 warnings`

lib/pyld/canon.py
```diff
@@ -504,57 +581,73 @@ def create_hash_to_related(self, id_, issuer):
...
+
+        if l_.language:
+            if l_.datatype:
+                raise Exception("Literal has datatype AND language!")
+            return f"{encoded}@{l_.language}"
+        elif l_.datatype and l_.datatype != XSD.string:
+            return f"{encoded}^^<{l_.datatype}>"
+        else:
+            return f"{encoded}"
+
```
gjhiggins, line 610: This looks like a good candidate for replacing the `_quote_encode` implementation in RDFLib's NTriples Serializer - [LoC#84](https://github.com/RDFLib/rdflib/blob/0fdac395078b81313f30eeab74b1a18eb7bb4db6/rdflib/plugins/serializers/nt.py#L84). I tried it out and all the RDFLib test suites pass. Might it be more direct to have `parts` as a string and append with `+=`? - it would at least obviate the ultimate `join`.

mielvds: Ah yes, ofc. Changed it. Thanks!

lib/pyld/canon.py
mielvds: Indeed. The test failures are because of the implicit `xsd:string` datatype I added. This was necessary to string compare test output, but is a good idea in general.
I opened two PR's at rdflib to get the discussion going:
- https://github.com/RDFLib/rdflib/pull/3433
- https://github.com/RDFLib/rdflib/pull/3431

lib/pyld/canon.py
```diff
@@ -14,44 +21,82 @@ def __init__(self):
...
+        # If outputMap option is set, return the map of blank node identifiers.
+        if options.get('outputMap'):
+            return dict(sorted(bnode_map.items(), key=lambda item: item[1]))
+
+        # 8) Return the normalized dataset.
+        if (
+            options.get('format') == 'application/n-quads'
+            or options.get('format') == 'application/nquads'
+        ):
```
anatoly-scherbakov, line 74:
```suggestion
        if options.get('format') in {
            'application/n-quads',
            'application/nquads',
        }:
```

anatoly-scherbakov (approved):

PR #248, opened 2026-04-10
Fix expansion with `@base`

anatoly-scherbakov:
- **Guard active `@base` use in `_expand_iri` when `base` is not None**
- **Add regression test for `@base` vs property term expansion**


davidlehn:
- Why are those dev libs now needed?  Did the github testing systems change behavior?
- If that test is not in the main JSON-LD API test suite, please add it there immediately.  The WG needs to agree on the behavior (I assume they will in this case) and have tests to ensure all implementations interoperate.  That also removes the need to have the test duplicated here.
- If the above can not be done now for some reason, then before merging this, please add tracking issues in the json-ld api repo and here to add the test case and clean it up here.

tests/test_base_vs_vocab.py
```diff
@@ -0,0 +1,32 @@
+"""
+Regression test: @base must not be used to expand property terms.
+
+Only @vocab is used to expand properties. @base only affects relative IRI
+resolution in values and @id fields. With only @base set (no @vocab), a plain
+term like `name` cannot be mapped to an absolute IRI and must be dropped from
+the expanded output.
+
+See: https://www.w3.org/TR/json-ld11/#iri-expansion (step 4 vs step 5 —
```
davidlehn: Looks like this should be https://www.w3.org/TR/json-ld11-api/#iri-expansion?

tests/test_base_vs_vocab.py
```diff
@@ -0,0 +1,32 @@
+"""
+Regression test: @base must not be used to expand property terms.
+
+Only @vocab is used to expand properties. @base only affects relative IRI
```
davidlehn: Needs rewording.  Regular terms work too, it's not "only" `@vocab`.

anatoly-scherbakov:
@davidlehn, thanks for the review!

* Updated the test comment as suggested :+1: 
* `libxml2-dev` / `libxslt1-dev` are needed when CI builds lxml from source on PyPy (no wheel for that combo). I did not check how it ran before, it failed for me, and I fixed it.

This bug is a very recent regression. I haven't found a test in the suite that would cover it. Which explains why the bug could appear.

PR in `json-ld-api` added: https://github.com/w3c/json-ld-api/pull/686 — but, it is unclear when that can be merged. For that reason, I propose we merge this PR. I do not consider a redundant test a problem, provided that the test is spec compliant.

lib/pyld/jsonld.py
```diff
@@ -5842,8 +5842,7 @@ def _expand_iri(
 
         # resolve against base
         rval = value
-        # if there is a base in the active context, use that
-        if '@base' in active_ctx:
+        if base is not None and '@base' in active_ctx:
```
mielvds, line 5845: Could you preserve the comment and adjust where appropriate?

tests/test_base_vs_vocab.py
```diff
@@ -0,0 +1,33 @@
+"""
+Regression test: @base must not be used to expand property keys.
```
mielvds, line 2: I'd prefer to have this as part of https://github.com/digitalbazaar/pyld/blob/9a9921c37f189961a1fc8d6ff9eb45e0743e0059/tests/test_jsonld.py#L10. I have a refactor of the class-based tests to function-based tests somewhere on the todo list, but for the time being, let's follow the same structure.

mielvds (approved):
Thanks for the fix @anatoly-scherbakov! Given that this doesn't break any other test, I guess this must of slipped through when introducing the default base URI. As long as the case is registered with the test-suite, I also prefer merging to preserve backwards-compatibility.

mielvds:
@anatoly-scherbakov I did a cleanup in master, so please rebase so my commits are no longer in your PR

anatoly-scherbakov:
Thanks for review! Comments addressed. cc @mielvds

mielvds (approved):
All good! Much appreciated

PR #250, opened 2026-04-26
Add FrozenDocumentLoader for offline / vetted JSON-LD contexts

anatoly-scherbakov:
## Why?

For tight security deployments, it might be necessary to:

* Ensure that the system does not require network access and does not download arbitrary data from the Web,
* But still make certain white listed contexts/remote documents available to JSON-LD processing.

## Summary

- Adds `pyld.DocumentLoader` (ABC) + `pyld.RemoteDocument` (`TypedDict`) — the first class-based loader contract in PyLD; existing function-based loaders remain valid.
- Adds `pyld.FrozenDocumentLoader`: a class-based loader that serves only URLs in its `documents` allowlist and refuses everything else with `JsonLdError(code='loading document failed')`. Suitable for air-gapped runs, reproducible builds, and security-hardened deployments. Honors the W3C *JSON-LD Best Practices* recommendation that clients SHOULD attempt to use a locally cached version of contexts ([§ Cache JSON-LD Contexts](https://w3c.github.io/json-ld-bp/#cache-json-ld-contexts)).
- Adds `pyld.BUNDLED_CONTEXTS`: 8 vendored W3C / W3ID JSON-LD contexts (ActivityStreams, DID v1, VC v1/v2, Linked Data Security v1/v2, Ed25519-2020, JWS-2020), refreshable with `make download-bundled-contexts`. With no arguments `FrozenDocumentLoader()` serves these; pass `dict(BUNDLED_CONTEXTS, **extras)` to extend.

### Design notes

- `documents: dict[str, dict | Path]` — `Path` entries are read & parsed lazily on first request and cached in place. `__post_init__` makes a defensive copy so the caller's mapping is never mutated.
- `RemoteDocument` is a real `TypedDict` (previously only a docstring word). Subclasses get a typed `__call__` return.
- Bundled `*.jsonld` files ship via `setup.py`'s `package_data` and `MANIFEST.in`'s `recursive-include`.

## Test plan

- [x] `pytest tests/test_frozen_document_loader.py -v` — 8 new tests, all pass.
- [x] Full non-network suite: `pytest -m "not network"` — 1477 passed, 43 skipped, 14 xfailed.
- [x] `make lint` clean.
- [x] End-to-end smoke: `jsonld.expand(doc_with_did_v1_context, options={'documentLoader': FrozenDocumentLoader()})` resolves with no network.
- [x] Refusal smoke: unknown URL raises `JsonLdError(code='loading document failed', type='jsonld.LoadDocumentError')`.
- [x] Bundle refresh: `make download-bundled-contexts` succeeds and is idempotent.

### DIFF
--- a/lib/pyld/documentloader/base.py
+++ b/lib/pyld/documentloader/base.py
@@ -1,0 +1,44 @@
+"""
+Abstract base class for class-based JSON-LD document loaders.
+
+.. module:: jsonld.documentloader.base
+  :synopsis: DocumentLoader abstract base class
+"""
+
+from abc import ABC, abstractmethod
+from typing import Any, TypedDict
+
+
+class RemoteDocument(TypedDict):
+    """Shape returned by a JSON-LD document loader.
+
+    Mirrors the *RemoteDocument* structure defined in the W3C JSON-LD 1.1 API
+    (https://www.w3.org/TR/json-ld11-api/#remotedocument).
+    """
+
+    contentType: str
+    contextUrl: str | None
+    documentUrl: str
+    document: Any
+
+
+class DocumentLoader(ABC):
+    """Abstract base class for class-based JSON-LD document loaders.
+
+    Concrete subclasses implement :meth:`__call__` to fetch a document for a
+    given URL and return a :class:`RemoteDocument`.
+
+    Existing function-based loaders (:func:`pyld.jsonld.requests_document_loader`,
+    :func:`pyld.jsonld.aiohttp_document_loader`) remain valid: pyld's loader
+    contract is "any callable with the right signature". This ABC is for new
+    class-based loaders only.
+    """
+
+    @abstractmethod
+    def __call__(self, url: str, options: dict) -> RemoteDocument:
+        """Retrieve the JSON-LD document at ``url``.
+
+        :param url: the URL to retrieve.
+        :param options: loader options (e.g. ``headers``).
+        :return: a :class:`RemoteDocument`.
+        """
--- a/lib/pyld/documentloader/frozen/__init__.py
+++ b/lib/pyld/documentloader/frozen/__init__.py
@@ -1,0 +1,85 @@
+"""
+Frozen JSON-LD document loader.
+
+A document loader that serves *only* the URLs in its ``documents`` allowlist
+and refuses everything else with :class:`pyld.jsonld.JsonLdError`. Suitable for
+secure / air-gapped / privacy-sensitive deployments and for honoring the
+guidance in the W3C *JSON-LD Best Practices* note that clients SHOULD attempt
+to use a locally cached version of contexts (§ Cache JSON-LD Contexts,
+https://w3c.github.io/json-ld-bp/#cache-json-ld-contexts).
+
+This module also defines :data:`BUNDLED_CONTEXTS`, a curated mapping of
+high-traffic public W3C / W3ID JSON-LD context URLs to vendored on-disk copies
+shipped with the package. See ``scripts/download_contexts.py`` for how the
+files in ``bundled/`` are refreshed.
+
+.. module:: jsonld.documentloader.frozen
+  :synopsis: FrozenDocumentLoader and BUNDLED_CONTEXTS
+"""
+
+import json
+from collections.abc import Mapping
+from dataclasses import dataclass, field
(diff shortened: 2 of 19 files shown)


mielvds:
Hi @anatoly-scherbakov , thanks for this. This is a useful feature. Before I review, two questions:
- it is necessary to have these local contexts in this repo? This seems difficult to maintain. Can't `make download-bundled-contexts` be part of the install procedure?
- do we need to revisit unittests that require network?

anatoly-scherbakov:
@mielvds thanks for the review!

> Can't `make download-bundled-contexts` be part of the install procedure?

I would advise against that.

* If `pyld` is being installed from a corporate PyPI mirror in a secure environment, Internet access might be restricted, and the installation will fail
* Or, network just might be glitchy

Developers who employ the new `FrozenDocumentLoader` _will_ vendor in the contexts, — that's the whole point of it. It is expected that vendored contexts are:

- vetted,
- verified for security,
- and will _not_ change.

This is the most straightforward way to avoid security vulnerabilities introduced by context spoofing.

For future versions of JSON-LD, the Working Group is going to work on security features that should reduce the possibility of context spoofing attacks.

Regardless of these developments, an air-gapped `DocumentLoader` that does not allow network access and whitelists remote documents is a solution already available to current JSON-LD systems.

cc @BigBlueHat what do you think?

lib/pyld/documentloader/frozen/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+    'https://w3id.org/security/v1': _BUNDLED_DIR / 'security-v1.jsonld',
+    'https://w3id.org/security/v2': _BUNDLED_DIR / 'security-v2.jsonld',
+    'https://w3id.org/security/suites/ed25519-2020/v1': _BUNDLED_DIR
+    / 'security-ed25519-2020-v1.jsonld',
+    'https://w3id.org/security/suites/jws-2020/v1': _BUNDLED_DIR
+    / 'security-jws-2020-v1.jsonld',
+}
+
+
+@dataclass
```
mielvds: Why is this a dataclass?

lib/pyld/documentloader/base.py
```diff
@@ -0,0 +1,44 @@
...
+    (https://www.w3.org/TR/json-ld11-api/#remotedocument).
+    """
+
+    contentType: str
+    contextUrl: str | None
+    documentUrl: str
+    document: Any
+
+
+class DocumentLoader(ABC):
```
mielvds, line 25: Can the original document loaders be refactored to use this?

tests/test_frozen_document_loader.py
```diff
@@ -0,0 +1,111 @@
...
+    FrozenDocumentLoader,
+    jsonld,
+)
+from pyld.jsonld import JsonLdError
+
+_URL = 'https://example.com/ctx'
+_INLINE_DOC = {'@context': {'name': 'http://schema.org/name'}}
+
+
+def test_loader_returns_remote_document_for_dict_entry():
```
mielvds, line 20: Can you add some docstring per test to give some more context on what it is testing?

mielvds (approved):

lib/pyld/documentloader/base.py
anatoly-scherbakov: Yes, and they should, but this would be out of scope for this particular PR.

lib/pyld/documentloader/frozen/__init__.py
anatoly-scherbakov: To avoid the need to write `__init__()`. Although, since we have to do `__post_init__()`, this might not be a good thing to do in this particular case. Thanks for flagging this, I'll check.

anatoly-scherbakov: Done

tests/test_frozen_document_loader.py
anatoly-scherbakov: Done

lib/pyld/documentloader/base.py
mielvds: understood and I agree

mielvds (approved):
Looks good to me! But let's give the original maintainers a couple of days to comment

Issue #247, opened 2026-04-05
Compaction term selection order changed in 3.0.0 (removal of _compare_shortest_least)


simontaurus:
In 2.0.4, _compare_shortest_least was used to deterministically sort candidate terms in the inverse context - shortest first, then lexicographically least. This function was removed in [24b1dd5](https://github.com/digitalbazaar/pyld/commit/24b1dd579dc01425e713e44b8d44fa421b122e33) ("Simplify compare functions for sorting"), which changes which term is selected when multiple context keys map to the same IRI during jsonld.compact().

This is a breaking change for anyone relying on a specific term being preferred over another when the context contains duplicate IRI mappings. It would be helpful to document this in the changelog or migration notes for 3.0.0.

davidlehn:
Do you have examples?  I thought there were spec test cases to catch some ordering issues.  Maybe there are some cases that are not covered?  Or this was specifically not a covered use case?  If there is some ordering behavior that should be deterministic, there certainly needs to be spec test coverage for that.

mielvds:
This was not intentional, so if some behavior changed, this is a bug. 
That said, if this ordering is not in the spec (not sure; no test picked this up), I'm hesitant to work on a fix, unless the code change would be minor. Happy to document it as a breaking in the changelog, though.

For my understanding: could you share a minimal example of what changed?

anlambert:
We also got it by the terms ordering change in `pyld 3.0` as some tests in our [swh-coarnotify project](https://gitlab.softwareheritage.org/swh/devel/swh-coarnotify) started [to fail](https://jenkins.softwareheritage.org/blue/organizations/jenkins/swh-coarnotify%2Fmaster/detail/master/421/pipeline) after bumping pyld.

Based on my understanding, commit https://github.com/digitalbazaar/pyld/commit/24b1dd579dc01425e713e44b8d44fa421b122e33 ensures the terms sorting respect the [JSON-LD specification](https://www.w3.org/TR/json-ld11-api/#overview-1) for inverse context creation: sort terms by length then sort terms having same length by lexicographical order. Previous ordering was not honoring JSON-LD specification as terms were sorted by lexicographical order then by length (so terms starting with an uppercase character were appearing before those starting with a lowercase character).

[...]

PR #252, opened 2026-05-02
Document compact term selection order change (#247)

simontaurus:
## Summary
- Add changelog entry documenting the compaction term selection order change as a **BREAKING** change in 3.0.0
- Add three unit tests to `TestCompact` verifying the spec-compliant behavior:
  - Shorter term is preferred over lexicographically-first longer term
  - Underscore-prefixed longer terms do not beat shorter terms
  - Same-length terms use lexicographic tiebreak

Addresses #247. The term ordering now matches the [W3C JSON-LD 1.1 API spec, section 4.3 step 3](https://w3c.github.io/json-ld-api/#inverse-context-creation): *"ordered by shortest term first (breaking ties by choosing the lexicographically least term)"*.

## Test plan
- [x] `pytest tests/test_jsonld.py::TestCompact` passes (5/5)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 ## 3.0.0 - 2026-04-02
 
 ### Changed
+- **BREAKING**: Compaction term selection now follows the JSON-LD 1.1 spec
+  (Inverse Context Creation, section 4.3 step 3): terms are ordered by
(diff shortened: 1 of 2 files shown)


mielvds (approved):
This is great, thanks!

PR #254, opened 2026-05-04
Migrate JsonLdProcessor to RDFLib

mielvds:
This PR migrates `JsonLdProcessor` to rdflib and completes the internal migration to rdflib.  It introduces a legacy mode to preserve backward-compatibility of the public API.

### Changed

- Migrated `jsonld.to_rdf()`, `jsonld.from_rdf()`, and N-Quads parsing/serialization to use `rdflib.Dataset` and RDFLib terms directly.
  - `jsonld.to_rdf()` now returns an `rdflib.Dataset` by default when no output `format` is requested.
  - Added `legacyMode` to `jsonld.to_rdf()` to return the previous RDF.js-like dataset `dict`.
  - `jsonld.from_rdf()` now accepts both `rdflib.Dataset` inputs and the previous legacy dataset `dict` shape.
- N-Quads input/output now delegates to RDFLib instead of PyLD’s custom parser and serializer.

### Removed

- Removed the internal `pyld.nquads` parser/serializer module.

### Fixed

- Preserved RDF literal lexical forms when converting through RDFLib, including canonical double output, large numeric values, and compound literal handling.
- Fixed `iri_resolver.unresolve()` query/fragment reconstruction while cleaning up the resolver docstring.

### Maintenance

- Updated RDF-related tests for RDFLib dataset and N-Quads serialization behavior.
- Removed small Python 2-era compatibility code and simplified a few internal collection checks.


mielvds:
@anatoly-scherbakov et al. would you mind doing an in-depth review of the changes to `jsonld.py`? The changes should be overseeable, but they are crucial to making the switch to 4.0 and rdflib.

anatoly-scherbakov (approved):

lib/pyld/jsonld.py
```diff
@@ -1381,27 +1370,52 @@ def parse_nquads(input_):
...
-        except ParserError as cause:
+            dataset = Dataset()
+            # TODO: workaround to preserve bnodes for testing and preserve previous behaviour;
+            # this should be handled in the test harness instead
+            bnode_context = {
+                label: label
+                for label in re.findall(r'_:([A-Za-z0-9_][A-Za-z0-9_.-]*)', input_)
+            }
+            parser = NQuadsParser()
+            rdflib.NORMALIZE_LITERALS = False
```
anatoly-scherbakov: This affects the global state of `rdflib` and therefore the global state of the whole application. Perhaps there could be a more localized way to do this?

tests/test_jsonld.py
```diff
@@ -946,27 +946,11 @@ def test_double_and_float_values(self):
             ],
         }
 
-        expected = {
```
anatoly-scherbakov, line 949: This removes `legacyMode` tests but perhaps they should be kept and cover `legacyMode` to ensure it continue working correctly while it is supported.

mielvds: Agreed that we should test this, but I think it's better to have a test_legacy_mode.py with a couple of dedicated interface tests? Additionally, I'd like to have tests for `util.from_legacy_dataset` and `util.to_legacy_dataset`

lib/pyld/jsonld.py
mielvds: I agree, but unfortunately, there's no easy way around this. Normalized literals cause a bunch of issues, such as https://github.com/RDFLib/rdflib/issues/3426 and some failing test-suite tests, and this is the only way to turn them off for the parser. 

I added subclass `util.UnnormalizedNQuadsParser`, which constructs rdflib Literals with `normalize=False` instead of mutating rdflib.NORMALIZE_LITERALS. This should eventually be part of rdflib though.

But like the TODO says, only the test harness uses this method, at least internally. With the integration of RDFLib, there's really no reason to keep `parse_nquads` besides backwards-compatibility, as one can use `NQuadsParser()` directly.

We could do the following:
- mark `parse_nquads` as deprecated in 3.2.0
- in 4.0.0
  - remove `parse_nquads`  OR 
  - keep `parse_nquads` with  to_legacy_dataset (no legacyMode) OR 
  - keep `parse_nquads` and the old parser

tests/test_jsonld.py
mielvds: I added two dedicated legacy mode tests; might expand and move those later.

PR #263, opened 2026-05-12
Fix scoped context and conflicting properties

### LINKED ISSUE #204: `@nest`ed node with own vocab has conflicting value overriden
From what I gather from the JSON-LD 1.1 specs, there is no problem providing its own `@context` to a `@nest`ed node. Yet, if the parent `@context` has a conflicting property name, it gets completely ignored in favor of the root vocab:

```py
from pyld import jsonld
import json
JSON = """ {
   "@context": {
    "dublinCore": {
      "@id": "http://foo.bar/dc",
      "@context": {
         "title": "http://purl.org/dc/terms/title"
      }
    },
    "title": "http://foo.bar/title"
   },
    "@id": "http://foo.bar/obj/test",
    "title": "test",
    "dublinCore": {
      "title": "Chapter 1: Jonathan Harker's Journal"
    }
}"""
doc = json.loads(JSON)
nquads = jsonld.to_rdf(doc, options={'format': 'application/n-quads'})
print(nquads)
```

prints 

```
<http://foo.bar/obj/test> <http://foo.bar/dc> _:b0 .
<http://foo.bar/obj/test> <http://foo.bar/title> "test" .
_:b0 <http://purl.org/dc/terms/title> "Chapter 1: Jonathan Harker's Journal" .
```

But nested does not care about the vocab

```py
from pyld import jsonld
import json

JSON = """ {
   "@context": {
    "dublinCore": {
      "@id": "@nest",
      "@context": {
         "title": "http://purl.org/dc/terms/title"
      }
    },
    "title": "http://foo.bar/title"
   },
    "@id": "http://foo.bar/obj/test",
    "title": "test",
    "dublinCore": {
      "title": "Chapter 1: Jonathan Harker's Journal"
    }
}"""
doc = json.loads(JSON)
nquads = jsonld.to_rdf(doc, options={'format': 'application/n-quads'})
print(nquads)
```

and prints

```
<http://foo.bar/obj/test> <http://foo.bar/title> "Chapter 1: Jonathan Harker's Journal" .
<http://foo.bar/obj/test> <http://foo.bar/title> "test" .

```

which seems wrong according to the JSON-LD 1.1 specs ?

[Edit: I edited to make the example as light as possible]

mielvds:
This PR fixes #204 and two related tests in the test-suite:

https://w3c.github.io/json-ld-api/tests/toRdf-manifest.html#tc037

> Nesting terms may have property-scoped contexts defined.

https://w3c.github.io/json-ld-api/tests/toRdf-manifest.html#tc038

> Nesting terms may have property-scoped contexts defined.

The issue was that `@nest` keys were delayed for later expansion before their property-scoped context was applied. This fix first gets the scoped `@context` from the `@nest` term and stores this property-scoped context with the key. The delayed nested object expansion then later uses that  property-scoped context, so for 

```json
{
  "@context": {
    "@vocab": "http://example.org/vocab#",
    "p1": {
      "@id": "@nest",
      "@context": {
        "p2": "http://example.org/ns#P2"
      }
    }
  },
  "p1": {
    "p2": "foo"
  }
}
```

`"p2"` expands to `http://example.org/ns#P2` instead of falling back to outer `@vocab`.


PonteIneptique:
The tests correspond to the initial issue :+1: Thanks for your work !

anatoly-scherbakov:
The changes might alter behavior in regard to nested contexts. I just pushed a test to this branch which aims to demonstrate that. I am sorry for pushing for a branch that's not mine, I will rather create a new, derived branch the next time @mielvds - I hope this won't create too much of an inconvenience.

mielvds:
> The changes might alter behavior in regard to nested contexts. I just pushed a test to this branch which aims to demonstrate that. I am sorry for pushing for a branch that's not mine, I will rather create a new, derived branch the next time @mielvds - I hope this won't create too much of an inconvenience.

Not at all @anatoly-scherbakov! Could you propagate this test to the JSON-LD test-suite? 🙏 

I moved some of the docstring around to where it fits best. Most of it made more sense in the actual code. I added an internal method `_prepare_nested_context` that handles these complex cases, which is basically moving the logic for finding the active type-scoped context from `_expand()` and calling it in the loop that I included in the original fix.

anatoly-scherbakov:
Propagated the nested scoped-context regression to the JSON-LD API test suite: https://github.com/w3c/json-ld-api/pull/689

mielvds:
@anatoly-scherbakov thanks! I did some final cleanup. It could still use a (minor) code review though.

anatoly-scherbakov (approved):

PR #266, opened 2026-05-17
[#261]: Refactor document loaders to subclasses

### LINKED ISSUE #261: Refactor document loaders to be subclasses of `DocumentLoader` base class
# Why

Presently, built in document loaders are function based.

# What

Derive them from `DocumentLoader` ABC for homogenuity.

API stays the same.

anatoly-scherbakov:
Closes #261

- Converts built-in requests and aiohttp document loaders to DocumentLoader subclasses while preserving factory call APIs.
- Adds factory tests for class-based loader instances.
- Ignores local .venv directories.

### DIFF
--- a/lib/pyld/documentloader/aiohttp.py
+++ b/lib/pyld/documentloader/aiohttp.py
@@ -14,6 +14,7 @@ import threading
 import urllib.parse as urllib_parse
 
 from pyld import iri_resolver
+from pyld.documentloader.base import DocumentLoader, RemoteDocument
 from pyld.jsonld import (JsonLdError, parse_link_header, LINK_HEADER_REL)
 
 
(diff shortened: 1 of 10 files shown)


lib/pyld/documentloader/aiohttp.py
```diff
@@ -147,11 +148,23 @@ def loader(url, options=None):
...
+            self.async_loader(url, headers), loop)
         return future.result()
 
-    return loader
+
+def aiohttp_document_loader(loop=None, secure=False, **kwargs):
+    """
+    Create an Asynchronous document loader using aiohttp.
+
+    :param loop: deprecated / ignored (kept for backward compatibility).
```
mielvds, line 163: if this is deprecated, maybe we can remove it from`AioHttpDocumentLoader`'s constructor?

mielvds (approved):
Thanks @anatoly-scherbakov ! Nothing much to add. I would still report this change in the changelog though.

anatoly-scherbakov:
@mielvds Thanks! Added to the changelog and to the docs.

Issue #259, opened 2026-05-07
More extensive documentation


mielvds:
This library has very limited documentation. A good documentation of the public API would be a start

PR #267, opened 2026-05-20
`cd docs/` → `make serve`

anatoly-scherbakov:
# Why

We might want to develop the docs a bit more, and for that purpose we need auto reloading on a dev stack.

# What

Add `make serve` command inside `docs/`.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -6,19 +6,22 @@ SPHINXOPTS    =
 SPHINXBUILD   = sphinx-build
 PAPER         =
 BUILDDIR      = _build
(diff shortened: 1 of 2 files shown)


mielvds (approved):
Wow this thing has docs 😄 I really haven't noticed this `docs` folder until now which is strange considering how much time I've spend with this repo lately.

Seems like they are ancient though. I've set #259 to 3.2.0, not sure we will manage. But all for already preparing for this.

anatoly-scherbakov:
@mielvds I think we should move docs to something more modern, and expand them. I can look into this in next PRs.

Issue #115, opened 2020-04-13
Add ContextResolver docs.


davidlehn:
The README needs ContextResolver docs.  Perhaps similar to:
https://github.com/digitalbazaar/pyld/pull/112#issuecomment-606897734

cc: @gkellogg @azaroth42

azaroth42:
That comment demonstrates how to customize it, but because the cache is generated in the module, it works by default without any changes to code (which is great!). 

(which is not to say the docs aren't needed, just that they should be clear that you only need to do that if you want to change the default behavior)

gkellogg:
Presumably, jsonld.js could use the same documentation.

BTW, I know @dlongley thought the `tag` aspect important; the only thing in either repo that would seem to be able to affect this is the `'tag': 'static'` done within the ContextResolver itself. Perhaps something got lost or is needed in the document loader, but it's not clear how or why this would be used.

yesteph:
Hi @gkellogg 

Indeed, the usage of `tag` fetched from a remote context is not clear.  No mention of this field in JSON-LD 1.0/1.1 specifications.

In our case, this leads to very poor performance. As the shared_cache is never populated we fetch remote contexts every times.

mielvds:
Added in #268 . Thanks @azaroth42 for the example 6 years ago ;)

PR #268, opened 2026-05-21
Add option to set MAX_CONTEXT_URLS in ContextResolver

### LINKED ISSUE #104: MAX_CONTEXT_URLS set to low
I have a context that is split over many context files (https://stuchalk.github.io/scidata/contexts/scidata.jsonld) and as a result is getting an error on the JSON-LD playground that uses this package.  Can the value be increased to 30, as 10 seems too low?

mielvds:
Fixes #104 

This would allow setting a custom ContextResolver with a different level, for instance 20:
```
contextresolver = ContextResolver(jsonld._resolved_context_cache, options['documentLoader'], max_context_urls=20)
jsonld.compact(input, context, {"contextResolver": contextresolver})
```

This is a ok fix, but the contextresolver could use a refactor. 
- The `cycles` parameter in `ContextResolver.resolve()` is a `set()`, but can also be an integer. 
- It feels like the default contextresolver should be part of the JsonLdProcessor object instead of being created in the method.
- according to the docs, "contextResolver" is "for internal use only", but I'm not sure why that is. @davidlehn @dlongley @BigBlueHat ?

### DIFF
--- a/lib/pyld/context_resolver.py
+++ b/lib/pyld/context_resolver.py
@@ -15,6 +15,7 @@ from pyld import iri_resolver, jsonld
 
 from .resolved_context import ResolvedContext
 
+# Restraints
 MAX_CONTEXT_URLS = 10
 
 
@@ -23,14 +24,28 @@ class ContextResolver:
     Resolves and caches remote contexts.
     """
 
-    def __init__(self, shared_cache, document_loader):
+    def __init__(self, shared_cache, document_loader, max_context_urls = None):
         """
         Creates a ContextResolver.
+
+        :param max_context_urls: the maximum number of times to recusively fetch contexts.
+          (default MAX_CONTEXT_URLS).
         """
         # processor-specific RDF parsers
         self.per_op_cache = {}
         self.shared_cache = shared_cache
         self.document_loader = document_loader
+        self._max_context_urls: int = max_context_urls if isinstance(max_context_urls, int) else MAX_CONTEXT_URLS
+
+    @property
+    def max_context_urls(self) -> int:
+        return self._max_context_urls
+
+    @max_context_urls.setter
+    def max_context_urls(self, max_context_urls: int) -> None:
+        if not isinstance(max_context_urls, int):
+            raise TypeError("Value for max_context_urls is not a number")
+        self._max_context_urls = max_context_urls
 
     def resolve(self, active_ctx, context, base, cycles=None):
         """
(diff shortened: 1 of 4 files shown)


lib/pyld/context_resolver.py
```diff
@@ -23,14 +24,28 @@ class ContextResolver:
     Resolves and caches remote contexts.
     """
 
-    def __init__(self, shared_cache, document_loader):
+    def __init__(self, shared_cache, document_loader, max_context_urls = None):
```
anatoly-scherbakov:
```suggestion
    def __init__(self, shared_cache, document_loader, max_context_urls: int = MAX_CONTEXT_URLS):
```

lib/pyld/context_resolver.py
```diff
@@ -23,14 +24,28 @@ class ContextResolver:
...
         Creates a ContextResolver.
+
+        :param max_context_urls: the maximum number of times to recusively fetch contexts.
+          (default MAX_CONTEXT_URLS).
         """
         # processor-specific RDF parsers
         self.per_op_cache = {}
         self.shared_cache = shared_cache
         self.document_loader = document_loader
+        self._max_context_urls: int = max_context_urls if isinstance(max_context_urls, int) else MAX_CONTEXT_URLS
```
anatoly-scherbakov:
```suggestion
        self.max_context_urls: int = max_context_urls
```

lib/pyld/context_resolver.py
```diff
@@ -23,14 +24,28 @@ class ContextResolver:
...
+
+    @property
+    def max_context_urls(self) -> int:
```
anatoly-scherbakov: Not sure why this is necessary. Perhaps just a property is enough?

anatoly-scherbakov (commented):
I believe `ContextResolver` is said to be internal only because JSON-LD spec does not mandates its existence. It mandates existence and API of `DocumentLoader` though.

lib/pyld/context_resolver.py
mielvds: I added this for future-proofing and also for making sure the value set is a number. But I agree this is not necessary at this stage and we can go for simple.

lib/pyld/context_resolver.py
mielvds: Shouldn't we safeguard this against random input? Typing doesnt get you far

anatoly-scherbakov: A piece of random input will legitimately fail when trying to use the invalid value, and the developer using the library will see that. Stack trace is a piece of useful debug info. Perhaps they will figure it out?

anatoly-scherbakov (approved):

mielvds:
I made it simpler and also added a README entry which also covers https://github.com/digitalbazaar/pyld/issues/115

PR #270, opened 2026-05-28
Implement compound-literal value conversion and add tests

mielvds:
This PR implements support for compound literals in RDF direction (see https://www.w3.org/TR/json-ld-api/#overview-16). It fixes the two related tests in the test-suite:

> Parses compound literal with proper option.

- [fromRdf#tdi11](https://w3c.github.io/json-ld-api/tests/fromRdf-manifest.html#tdi11) 
- [fromRdf#tdi12](https://w3c.github.io/json-ld-api/tests/fromRdf-manifest.html#tdi12)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,10 @@
   which replaces the role of the static `MAX_CONTEXT_URLS`. The constructor 
   now accepts a `max_context_urls` parameter that sets the value of 
   `max_context_urls` which defaults to `MAX_CONTEXT_URLS`. 
-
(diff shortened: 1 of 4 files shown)


mielvds:
@anatoly-scherbakov I merged your tests and accounted for these cases. Also, I added one for multiple `@value` entries. 

- It now first decodes all compound literals and then looks for references. When references are shared, the object is not nested.
- It checks for invalid compositions of `rdf:direction`, `rdf:language` and `rdf:value` before turning them into objects. Else it raises.

anatoly-scherbakov (approved):

anatoly-scherbakov (approved):

PR #274, opened 2026-06-06
Add compound literal from RDF regression tests

anatoly-scherbakov:
- shared compound-literal blank nodes must remain ordinary nodes when referenced more than once
- invalid rdf:direction values must raise invalid base direction
- invalid rdf:language values must raise invalid language-tagged string

These tests currently fail against the implementation in #270 and are intended to document the expected behavior before the implementation is adjusted.


mielvds:
Thanks @anatoly-scherbakov, cases are legit. They should be part of the test-suite, at least the first one.

PR #291, opened 2026-07-01
[#253]: Implement `SqliteCacheRequestsDocumentLoader`

### LINKED ISSUE #253: Implement disk-based cache for remote documents
W3C *JSON-LD Best Practices* §[Cache JSON-LD Contexts](https://w3c.github.io/json-ld-bp/#cache-json-ld-contexts):

> Services providing a JSON-LD Context SHOULD set HTTP cache-control headers to allow liberal caching of such contexts, and clients SHOULD attempt to use a locally cached version of these documents.

The bundled function loaders re-fetch every URL on every call, and [`_resolved_context_cache`](https://github.com/digitalbazaar/pyld/blob/cba2e29cad80472dcfd4a5bf654019bf95224550/lib/pyld/jsonld.py#L135-L136) is in-process only — so short-lived processes (CLIs, workers, serverless) re-download every context on every run. See #70.

### Proposal

Add a `DocumentLoader` subclass (the ABC introduced in #250) that persists fetched documents to a configurable on-disk cache directory and honors `Cache-Control` / `Expires` / `ETag` / `Last-Modified` for revalidation.

```python
@dataclass
class CachingDocumentLoader(DocumentLoader):
    cache_dir: Path
    default_max_age: int | None = None
```

Opt-in; existing loaders unchanged. Async sibling can follow.

Related: #70, #85, #250.

anatoly-scherbakov:
# Why

We need to provide a method of persistent disk based HTTP-aware caching.

# What

Here, we are discussing our options and implementing one of them.

### DIFF
--- a/docs/project/decisions/use-requests-cache-for-sync-http-caching-in-document-loaders.md
+++ b/docs/project/decisions/use-requests-cache-for-sync-http-caching-in-document-loaders.md
@@ -1,0 +1,183 @@
+---
+title: Use requests-cache for persistent HTTP caching in synchronous Python code
+status: decided
+date: 2026-06-29
+author: Anatoly Scherbakov
+tags: [decision]
+hide: [toc]
+---
+
+# Use `requests-cache` for persistent HTTP caching in synchronous Python code
+
+{{ adr_metadata(date, status) }}
+
+## :material-text-box-outline: Context
+
+[:material-file-download-outline: PyLD document loaders](/pyld/reference/document-loaders/) fetch remote JSON-LD contexts and other documents over HTTP. [JSON-LD Best Practices](https://w3c.github.io/json-ld-bp/) recommends:
+
+> **Best Practice 14:** Cache JSON-LD Contexts
+>
+> Services providing a JSON-LD Context *SHOULD* set HTTP cache-control headers to allow liberal caching of such contexts, and clients *SHOULD* attempt to use a locally cached version of these documents.
+>
+> — [§ 8.1 Cache JSON-LD Contexts](https://w3c.github.io/json-ld-bp/#cache-json-ld-contexts)
+
+### Status Quo
+
+PyLD already caches remote contexts in process memory only. [`ContextResolver`](https://github.com/digitalbazaar/pyld/blob/master/lib/pyld/context_resolver.py) shares a module-level [`LRUCache`](https://github.com/digitalbazaar/pyld/blob/master/lib/pyld/jsonld.py#L160-L162) of resolved contexts, and [`ResolvedContext`](https://github.com/digitalbazaar/pyld/blob/master/lib/pyld/resolved_context.py) keeps a per-document [`LRUCache`](https://github.com/digitalbazaar/pyld/blob/master/lib/pyld/resolved_context.py#L28) for processed contexts relative to an active context.
+
+[`FrozenDocumentLoader`](/pyld/reference/document-loaders/frozen/) serves files from disk based on its configuration options set in code.
+
+None of this is HTTP-aware or persistent across processes: [`RequestsDocumentLoader`](/pyld/reference/document-loaders/requests/) and [`AioHttpDocumentLoader`](/pyld/reference/document-loaders/aiohttp/) issue a fresh HTTP request on every load with no `Cache-Control` handling.
+
+### HTTP Caching Standards
+
+Client-side caching semantics are governed by the following standards:
+
+```mermaid
+flowchart LR
+  rfc2616("<b>RFC 2616</b><br/>HTTP/1.1")
+  rfc7234("<b>RFC 7234</b><br/>HTTP/1.1: Caching")
+  rfc9111("<b>RFC 9111</b><br/>HTTP Caching")
+  rfc9110("<b>RFC 9110</b><br/>HTTP Semantics")
+  rfc5861("<b>RFC 5861</b><br/>Cache-Control Extensions for Stale Content")
+  rfc8246("<b>RFC 8246</b><br/>HTTP Immutable Responses")
+
+  rfc2616 -->|is superseded by| rfc7234
+  rfc7234 -->|is superseded by| rfc9111
+  rfc9110 -->|complements| rfc9111
+  rfc5861 -->|complements| rfc9111
+  rfc8246 -->|complements| rfc9111
+
+  click rfc2616 "https://www.rfc-editor.org/rfc/rfc2616.html" "RFC 2616"
+  click rfc7234 "https://www.rfc-editor.org/rfc/rfc7234.html" "RFC 7234"
+  click rfc9111 "https://www.rfc-editor.org/rfc/rfc9111.html" "RFC 9111"
+  click rfc9110 "https://www.rfc-editor.org/rfc/rfc9110.html" "RFC 9110"
+  click rfc5861 "https://www.rfc-editor.org/rfc/rfc5861.html" "RFC 5861"
+  click rfc8246 "https://www.rfc-editor.org/rfc/rfc8246.html" "RFC 8246"
+```
+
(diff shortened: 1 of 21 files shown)


docs/project/decisions/use-requests-cache-for-sync-http-caching-in-document-loaders.md
```diff
@@ -0,0 +1,183 @@
+---
```
mielvds, line 1: I'm on the fence about the added-value. It documents things, which is always good, but it also adds bloat (which is what AI agents are really good at :)) that will be hard to maintain? Also, there wasn't really any formal decision process to begin with. That said, it also doesn't hurt having them.

Personally, I prefer this summarized as something more usage-oriented in the docs (for instance as https://squidfunk.github.io/mkdocs-material/reference/admonitions/ to give background on the decision), the README, and the changelog, which is - looking at the rest of this PR - more or less the case. What do you think @anatoly-scherbakov? Is this something you do in other projects?

lib/pyld/documentloader/requests.py
```diff
@@ -15,16 +15,18 @@
...
+from pyld.jsonld import LINK_HEADER_REL, JsonLdError, parse_link_header
 
 
 class RequestsDocumentLoader(DocumentLoader):
     """Remote document loader using Requests."""
 
-    def __init__(self, secure=False, **kwargs):
-        import requests
+    def __init__(self, secure=False, session=None, **kwargs):
+        if session is None:
```
mielvds, line 25: This is necessary for the caching to work?

lib/pyld/documentloader/requests_sqlite_cache.py
```diff
@@ -0,0 +1,55 @@
+"""
+SQLite-backed HTTP cache document loader using requests-cache.
```
mielvds, line 2: This is great, of course, but I wonder if we should have a simple/naive zero-dependency alternative as well for the smaller use cases?

lib/pyld/documentloader/requests_sqlite_cache.py
```diff
@@ -0,0 +1,55 @@
...
+from pyld.documentloader.base import DocumentLoader, RemoteDocument
+from pyld.documentloader.requests import RequestsDocumentLoader
+
+
+def _resolve_sqlite_file_path(sqlite_file_path: Path | None) -> Path:
+    """Return absolute path to the SQLite cache file."""
+    from platformdirs import user_cache_dir
+
+    if sqlite_file_path is None:
+        return Path(user_cache_dir('pyld')) / 'http_cache.sqlite'
```
mielvds: Should Path not be used to resolve to whole path?

mielvds (approved):
Thanks @anatoly-scherbakov, very useful addition 👍

docs/project/decisions/use-requests-cache-for-sync-http-caching-in-document-loaders.md
anatoly-scherbakov: I have been using ADR (Architecture Decision Record) format before, but in a rather limited fashion. I always wanted to do it with higher quality so that's what I tried here; I think this document is the most advanced ADR I've ever done :)

The idea of an ADR is not just to document a decision but also to ensure that the quality of that decision is as high as possible. For example, here we are doing a conscious effort to include all alternatives and to  take in account everything that matters about them, including their popularity (and, therefore, their supporting communities).

That's the reason of the bloat (this document is large indeed). However, the rendered page seems rather condensed. Its largest part in code is a table which I tried to do as visually compact as I could.

I feel that making decisions _very_ explicit might benefit open source projects, surfacing challenges that might otherwise slip unnoticed.

As for maintenance, - the table does have "last updated date" and the decision document itself is also timestamped. Idea of an ADR document is that, when it is marked as Decided, it is no longer mutable. Bug fixes or typos can be remedied, but the substance of the document remains read-only forever.

If the nature of the decision has to be changed at any point, the document can and should be marked by Superseded by another decision.

What do you think?

lib/pyld/documentloader/requests.py
anatoly-scherbakov: Yes, basically, `requests-cache` library exposes a custom version of `requests.Session`; so, the ability to pass a session here enables us to use caching. We pass a special session to the default document loader.

It  is also possible for the user to configure a Session as they see fit and pass it. This can, say, enable in-memory caching.

lib/pyld/documentloader/requests_sqlite_cache.py
anatoly-scherbakov: We'd have to tackle multiple concerns:

- Filesystem locks and multiprocess behavior,
- HTTP caching semantics itself with its multitude of headers and rules,

so not sure actually. Not sure which subset of these features we could implement without noticeably increasing our maintenance burden. I agree that an additional dependency might be a hassle, but a well maintained library with its own community should be better long term than our ad hoc module.

anatoly-scherbakov (commented):
Thanks for the review @mielvds ! Left a few replies.

docs/project/decisions/use-requests-cache-for-sync-http-caching-in-document-loaders.md
mielvds: That's reasonable, so let's try it :)
I still have some questions on the process: when and what gets/needs an ADR? Does this need to be signed off by several maintainers or is some PR discussion enough to produce one?

lib/pyld/documentloader/requests_sqlite_cache.py
mielvds: Fair points. Originally, I thought you meant a generic wrapper that caches `__call__`, but anything else than HTTP caching semantics would also raise a bunch of implementation questions like cache invalidation. And users can easily implement such wrapper themselves; maybe we should add a short example somewhere on how to do that. There's also https://pypi.org/project/aiohttp-client-cache/, so we can add a similar implementation for AiHttp to the todo list.

docs/project/decisions/use-requests-cache-for-sync-http-caching-in-document-loaders.md
anatoly-scherbakov: I believe a PR discussion, like this one, is fine. If you agree with this decision as proposed, and this PR is merged, then the decision will land in master with the status of **Decided**. That should be enough.

If we can't reach an agreement, then it can land as **Draft** and we can refine it in further PRs.

If we retire it at some point we'll call it **Superseded** and write some other decision.

As to the cases where ADR is necessary at all, - I think that's determined by the consequences of the decision. If those consequences are sufficiently major then it might be a good idea to document them. For instance, in this case, we are discussing whether we do HTTP caching in the codebase or we introduce a dependency; that has consequences for the future, so that's why we document it.

lib/pyld/documentloader/requests_sqlite_cache.py
anatoly-scherbakov: `aiohttp-client-cache` is mentioned in the ADR, and we might want to implement it along with `requests-cache`, but I've left the corresponding column in the table Undecided, meaning, out of scope of this ADR. We can look into it later.

Here's the existing page about making a custom Document Loader: https://digitalbazaar.github.io/pyld/reference/document-loaders/custom/ - wdyt?

mielvds: Ah yes that perfectly sufficient, sorry for not keeping track 😅

docs/project/decisions/use-requests-cache-for-sync-http-caching-in-document-loaders.md
mielvds: Great, so let's add a clean version of your reply to https://github.com/digitalbazaar/pyld/blob/5954335169546546b461d62d0cf3e7b7504d776b/docs/project/index.md under a section 'decision process' and then we're good to go!

PR #297, opened 2026-07-09
Represent large numbers with no fractions as double for values >= 1e21

mielvds:
Fixes [toRdf-manifest#trt01](https://w3c.github.io/json-ld-api/tests/toRdf-manifest#trt01)

### DIFF
--- a/lib/pyld/jsonld.py
+++ b/lib/pyld/jsonld.py
@@ -3959,6 +3959,10 @@ class JsonLdProcessor:
                     object['value'] = _canonicalize_double(float_value)
                     object['datatype'] = XSD_DOUBLE
                     return object
+            # Numbers without fractions but larger or equal to 1e21 should be double
+            elif _is_integer(value) and value >= 1e21:
+                object['value'] = _canonicalize_double(value)
+                object['datatype'] = XSD_DOUBLE
             elif _is_integer(value):
                 object['value'] = str(int(value))
                 object['datatype'] = datatype or XSD_INTEGER
--- a/lib/pyld/jsonld.py
+++ b/lib/pyld/jsonld.py
@@ -3960,9 +3960,9 @@ class JsonLdProcessor:
                     object['datatype'] = XSD_DOUBLE
                     return object
             # Numbers without fractions but larger or equal to 1e21 should be double
-            elif _is_integer(value) and value >= 1e21:
+            elif _is_integer(value) and abs(value) >= 1e21:
                 object['value'] = _canonicalize_double(value)
-                object['datatype'] = XSD_DOUBLE
+                object['datatype'] = datatype or XSD_DOUBLE
             elif _is_integer(value):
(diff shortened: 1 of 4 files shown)


lib/pyld/jsonld.py
```diff
@@ -3959,6 +3959,10 @@ def _object_to_rdf(self, item, issuer, triples, rdf_direction):
                     object['value'] = _canonicalize_double(float_value)
                     object['datatype'] = XSD_DOUBLE
                     return object
+            # Numbers without fractions but larger or equal to 1e21 should be double
+            elif _is_integer(value) and value >= 1e21:
```
anatoly-scherbakov: Can this rule be gated to JSON-LD 1.1 processing? The `10^21` threshold is a JSON-LD 1.1 change. With `processingMode="json-ld-1.0"`, this unconditional branch changes a large integral value from the JSON-LD 1.0 integer representation to `xsd:double`. A regression test for 1.0 mode would help protect the existing behavior.

lib/pyld/jsonld.py
```diff
@@ -3959,6 +3959,10 @@ def _object_to_rdf(self, item, issuer, triples, rdf_direction):
...
+                object['value'] = _canonicalize_double(value)
+                object['datatype'] = XSD_DOUBLE
```
anatoly-scherbakov: Maybe we could preserve an explicitly supplied datatype here, as the surrounding branches do. The algorithm only sets the datatype to `xsd:double` when it is null; this assignment currently rewrites values explicitly typed as `xsd:integer` or with a custom datatype. This should likely be `datatype or XSD_DOUBLE`.

lib/pyld/jsonld.py
anatoly-scherbakov: Could we use `abs(value) >= 1e21` here? The JSON-LD 1.1 Object to RDF Conversion algorithm applies this threshold to the absolute value. As written, `-10**21` is serialized as `"-1000000000000000000000"^^xsd:integer`, but it should be `"-1.0E21"^^xsd:double`.

lib/pyld/jsonld.py
mielvds: sure, but if a custom datatype is supplied, should the value still be turned into a canonicalized double?

lib/pyld/jsonld.py
mielvds: I added a gate and a test

mielvds: Added `abs(value)`

lib/pyld/jsonld.py
anatoly-scherbakov: Yes indeed. Makes sense.

anatoly-scherbakov (approved):

lib/pyld/jsonld.py
mielvds: it makes sense to canonicalize the double?

PR #299, opened 2026-07-19
Add YAML-LD test suite & decide about YAML-LD implementation in `pyld`

anatoly-scherbakov:
# Why

[YAML-LD](https://www.w3.org/TR/yaml-ld-10/) is an emerging specification being developed by the [JSON-LD Working Group](https://www.w3.org/groups/wg/json-ld/). Normatively, it is quite stable already even though it is an [Editor's Draft](https://w3c.github.io/yaml-ld/) at the time of writing this.

I had already started a Python implementation for it at https://python-yaml-ld.iolanta.tech ([`iolanta-tech/python-yaml-ld`](https://github.com/iolanta-tech/python-yaml-ld)), `pip install yaml-ld`. It is basically a wrapper on top of [`pyld`](https://github.com/digitalbazaar/pyld).

It has to implement its own API functions such as `expand()` which call Respective `pyld` functions adding YAML support via custom Document Loaders.

I wonder if we should instead implement this as a feature of PyLD itself, guarded by an optional dependency:

```
pip install pyld[yaml-ld]
```

and make it so that `expand()` and other functions in `pyld` support YAML-LD. What do you guys think?

See also the undecided ADR in this PR: [`docs/project/decisions/choose-where-to-host-yaml-ld-support.md`](https://github.com/digitalbazaar/pyld/blob/yaml-ld-tests/docs/project/decisions/choose-where-to-host-yaml-ld-support.md).

As a follow up, we could think of implementing similar formats, such as:

* [CBOR-LD](https://w3c.github.io/cbor-ld/), which is being developed by the WG as well;
* Markdown-LD, which is a colloquial name for Markdown documents with YAML-LD in their frontmatters;
* maybe TOML-LD and many more formats from JSON-LD ecosystem.


mielvds:
@anatoly-scherbakov I see value in this, at least for yaml-ld, so fine by me. However, I don't want to have anything merged before 4.0.0 if that's ok.

PR #303, opened 2026-07-26
Defer SqliteCacheRequestsDocumentLoader cache file creation until first load

anatoly-scherbakov:
## Summary
- Create the SQLite HTTP cache file when the first document is loaded, not when `SqliteCacheRequestsDocumentLoader` is constructed.
- Keeps local-only CLI / library use from touching the cache path until an HTTP load happens.

## Stack
Base: `choice-by-type-document-loader` (#304).
Next: `command-line-interface` (#305).

## Test plan
- [x] `pytest tests/test_sqlite_cache_requests_document_loader.py`
- [ ] CI green


mielvds (approved):
Looks good! Pretty elegant improvement

PR #302, opened 2026-07-26
Add ChoiceBySchemeDocumentLoader for scheme-based document loading

anatoly-scherbakov:
## Summary
- Add `ChoiceBySchemeDocumentLoader`, which dispatches to per-scheme loaders so a local `file:` document can resolve a remote `https:` `@context`
- Export the loader, cover dispatch/fallback/errors with tests, and document it with an offline composed-loader example
- Add an `example_data` macro and tabbed sourced JSON-LD on the ChoiceByScheme and File loader pages

Stacked on #301 (`file-document-loader`). Closes #300.

## Test plan
- [ ] `pytest tests/test_choice_by_scheme_document_loader.py`
- [ ] `make test`
- [ ] `make lint`
- [ ] `make docs-build`
- [ ] Confirm docs page tabs show Example + sourced JSON-LD for ChoiceByScheme and File loaders

PR #301, opened 2026-07-25
Add FileDocumentLoader for local file: URLs

anatoly-scherbakov:
## Summary
- Add `FileDocumentLoader` to load local JSON-LD from `file:` URLs, scheme-less paths, and `pathlib.Path`, with optional `root` confinement
- Document the loader (reference page, examples, content-type table macro) and export it from the package root
- Switch `FrozenDocumentLoader` docs to mkdocstrings autodoc and clean Sphinx roles from its docstrings for MkDocs rendering

## Test plan
- [x] `pytest tests/test_file_document_loader.py`
- [x] `make docs-build` (strict)
- [ ] CI green on the PR

Related: prerequisite for https://github.com/digitalbazaar/pyld/issues/151; scheme dispatch tracked in https://github.com/digitalbazaar/pyld/issues/300

PR #304, opened 2026-07-26
Add `TypeDirectedDocumentLoader` for type-based document loading

anatoly-scherbakov:
## Summary
- Add `ChoiceByTypeDocumentLoader` to dispatch by Python input type (`pathlib.Path` vs URL `str`).
- Document and export the loader; compose with `ChoiceBySchemeDocumentLoader` for URL schemes.

## Stack
`master` → #301 File → #302 Scheme → **#304 ByType** (this PR) → #303 Sqlite lazy-init → #305 CLI.

## Test plan
- [x] `pytest tests/test_choice_by_type_document_loader.py tests/test_choice_by_scheme_document_loader.py`
- [ ] CI green

### DIFF
--- a/docs/examples/document_loaders/type_directed.py
+++ b/docs/examples/document_loaders/type_directed.py
@@ -1,0 +1,42 @@
+import json
+from pathlib import Path
+
+from pyld import (
+    FileDocumentLoader,
+    FrozenDocumentLoader,
+    SchemeDirectedDocumentLoader,
+    TypeDirectedDocumentLoader,
+    jsonld,
+)
+
+person = (
+    Path(__file__).resolve().parent.parent / 'data' / 'person_remote_context.jsonld'
+)
+
+file_loader = FileDocumentLoader()
+http_loader = FrozenDocumentLoader(
+    documents={
+        'https://example.com/context': {
+            '@context': {'name': 'http://schema.org/name'},
+        },
+    }
+)
+loader = TypeDirectedDocumentLoader(
+    {
+        Path: file_loader,
+        str: SchemeDirectedDocumentLoader(
+            file=file_loader,
+            http=http_loader,
+            https=http_loader,
+        ),
+    }
+)
+remote = jsonld.load_document(person, options={'documentLoader': loader})
+result = jsonld.expand(
+    remote['document'],
+    options={
+        'documentLoader': loader,
+        'base': remote['documentUrl'],
+    },
+)
+print(json.dumps(result, indent=2))
--- a/docs/reference/document-loaders/type-directed.md
+++ b/docs/reference/document-loaders/type-directed.md
@@ -12,7 +12,9 @@ hide: [toc]
 
 `TypeDirectedDocumentLoader` dispatches a value to the loader registered for its Python type. Applications can register any types that represent distinct document locations or loading policies.
 
-This example registers `Path` for local documents and `str` for URL values. A local document is loaded through `FileDocumentLoader`; its remote `@context` is a URL string, so the `str` registration delegates to a nested `SchemeDirectedDocumentLoader`.
+## Example use case
+
+One use case for `TypeDirectedDocumentLoader` is an application that accepts local `pathlib.Path` values and remote URL strings in the same JSON-LD workflow. This example routes paths through `FileDocumentLoader`; when a local document references a remote `@context`, its `str` URL is delegated to `SchemeDirectedDocumentLoader`.
 
 === "Example"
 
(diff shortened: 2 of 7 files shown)


anatoly-scherbakov:
Renamed to `TypeDirectedDocumentLoader` to match the previous one cc @mielvds

docs/examples/document_loaders/type_directed.py
```diff
@@ -0,0 +1,38 @@
...
+            'document': url.read_text(encoding='utf-8'),
+        }
+
+
+class StrDocumentLoader(DocumentLoader):
+    def __call__(self, url, options=None):
+        raise AssertionError(f'unexpected str load: {url}')
+
+
+loader = TypeDirectedDocumentLoader(
```
mielvds, line 24: I don't find this example very clear, but mostly because of the loader implementations. Perhaps it should just direct `Path` to `FileDocumentLoader` and `str` to `RequestsDocumentLoader` and some commentary line that very briefly explains what will happen.

docs/reference/document-loaders/type-directed.md
```diff
@@ -0,0 +1,21 @@
...
+# :material-shape-outline: `TypeDirectedDocumentLoader`
+
+::: pyld.TypeDirectedDocumentLoader
+    options:
+      show_root_heading: false
+      show_bases: false
+      heading_level: 3
+      members: false
+
+Dispatch by Python input type — here a `pathlib.Path` vs a URL `str`:
```
mielvds: Could you add a use case + example on when and how you would use this?

lib/pyld/jsonld.py
```diff
@@ -2763,8 +2763,7 @@ def _expand_object(
                 )
 
                 if any(
-                    self._expand_iri(active_ctx, k, vocab=True) == '@value'
-                    for k in nv
+                    self._expand_iri(active_ctx, k, vocab=True) == '@value' for k in nv
```
mielvds: Could you revert these formatting changes? This could be annoying for rebasing the PR's that are already open

tests/test_type_directed_document_loader.py
```diff
@@ -0,0 +1,116 @@
...
+    assert exc.value.details['types'] == ['Path']
+
+
+def test_load_path_with_remote_context(tmp_path):
+    """Load a Path whose @context is an https URL via composed loaders."""
+    path = tmp_path / 'person.jsonld'
+    path.write_text(json.dumps(_PERSON), encoding='utf-8')
+    file_loader = FileDocumentLoader()
+    http = FrozenDocumentLoader(documents={_CONTEXT_URL: _CONTEXT})
+    loader = TypeDirectedDocumentLoader(
```
mielvds, line 96: It would be cool to have a paragraph in the docs that illustrates such nesting scenario

mielvds (changes requested):
Neat! Just some small requests ;)

docs/reference/document-loaders/type-directed.md
anatoly-scherbakov: If my memory serves me, I think I used this in CLI implementation. This makes Python code a little bit more idiomatic if I may say so: we use one type for paths, another for URLs.

docs/examples/document_loaders/type_directed.py
anatoly-scherbakov: Reworked this to use the standard loaders: `Path` routes to `FileDocumentLoader`, while `str` enters `SchemeDirectedDocumentLoader`. That makes the composition explicit and lets the same example exercise the nested route; `FrozenDocumentLoader` keeps the rendered documentation deterministic.

tests/test_type_directed_document_loader.py
anatoly-scherbakov: Added an end-to-end nested example: the local `Path` document routes to `FileDocumentLoader`, and its HTTPS `@context` string routes through `SchemeDirectedDocumentLoader` to `FrozenDocumentLoader`.

docs/reference/document-loaders/type-directed.md
mielvds: Understood! But I mostly meant: can you add an use case + example in the docs for developers/users that are looking for a solution like this

anatoly-scherbakov: Thanks — I added an explicit [example use case](https://github.com/digitalbazaar/pyld/blob/16c10e95a57c48773236232da8481e32db59302a/docs/reference/document-loaders/type-directed.md#example-use-case) to the documentation. It shows one way an application can route local `pathlib.Path` values and remote URL strings through the same JSON-LD workflow, and ties that use case to the runnable nested-loader example.

PR #305, opened 2026-07-26
Add pyld command-line interface

### LINKED ISSUE #151: command-line tool
It would be great to have a command-line tool based on PyLD, for doing basing JSON-LD transformations.
So I hacked one for my own purposes. It is here:
    https://gist.github.com/pchampin/9416e423efe9925adaee536a3abdaf4c

Feel free to include it if you think it is a good idea.

anatoly-scherbakov:
## Summary
- Add `pyld` CLI (`get`, `expand`, `cache clear`) via `pip install "PyLD[cli]"`.
- Lazy console-script entry (`cli_entry`) so a bare install fails with an install hint instead of an import crash.
- Compose `ChoiceByTypeDocumentLoader` + `ChoiceBySchemeDocumentLoader` + cached HTTP loader for Path vs URL inputs.
- CLI docs (termynal examples), README/installation notes.

## Stack
`master` → #301 File → #302 Scheme → #304 ByType → #303 Sqlite lazy-init → **#305 CLI** (this PR).

## Test plan
- [x] `pytest tests/test_cli.py`
- [ ] CI green
- [ ] `make docs-build` after docs deps install

### DIFF
--- a/lib/pyld/cli/__init__.py
+++ b/lib/pyld/cli/__init__.py
@@ -1,0 +1,240 @@
+"""Command-line interface for PyLD."""
+
+from __future__ import annotations
+
+import json
+import shlex
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Annotated, Any
+from urllib.parse import urlparse
+
+try:
+    import typer
+except ImportError as cause:  # pragma: no cover
+    raise ImportError(
+        'The pyld command-line interface requires Typer. '
+        'Install it with: pip install "PyLD[cli]"'
+    ) from cause
+
+from rich.console import Console
+from rich.errors import NotRenderableError
+from rich.panel import Panel
+from rich.syntax import Syntax
+from typer import Argument, Option
+
+from pyld import (
+    FileDocumentLoader,
+    SchemeDirectedDocumentLoader,
+    SqliteCacheRequestsDocumentLoader,
+    TypeDirectedDocumentLoader,
+    jsonld,
+)
+
+
+@dataclass
+class State:
+    """Per-invocation CLI state carried on ``typer.Context.obj``."""
+
+    traceback: bool = False
+    cache_file: Path | None = None
+
+
+pyld = typer.Typer(
+    help='Command-line tool for JSON-LD transformations.',
+    no_args_is_help=True,
+)
+cache_app = typer.Typer(help='Cache management.', no_args_is_help=True)
+pyld.add_typer(cache_app, name='cache')
+pyld.state = State()  # type: ignore[attr-defined]
+
+console = Console()
+err_console = Console(stderr=True)
+
+MaybeStr = str | None
+
+InputArgument = Annotated[
+    str,
+    Argument(
+        metavar='INPUT',
+        help='Path or URL. Pass - to read from standard input.',
+    ),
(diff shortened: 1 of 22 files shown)


lib/pyld/cli/__init__.py
```diff
@@ -0,0 +1,239 @@
+"""Command-line interface for PyLD."""
```
mielvds, line 1: Can we put `cli.py` and `cli_entry.py` in a `cli/` subdirectory?

lib/pyld/cli.py
```diff
@@ -0,0 +1,239 @@
...
+    if scheme and len(scheme) != 1:
+        return value
+    return Path(value).expanduser().resolve()
+
+
+def is_stdin(value: MaybeStr) -> bool:
+    return value is None or value == '-'
+
+
+def as_document_url(location: str | Path) -> str:
```
mielvds: this is only used in combination with `parse_location`. Maybe a `document_url(value: str)` that calls `parse_location` is cleaner?

lib/pyld/cli/__init__.py
```diff
@@ -0,0 +1,239 @@
...
+    state = State(traceback=traceback, cache_file=cache_file)
+    ctx.obj = state
+    # Entry-point wrapper reads this after Click has torn down the context.
+    pyld.state = state  # type: ignore[attr-defined]
+
+
+@pyld.command('get')
+def get_command(input_: InputArgument = None) -> None:
+    """Retrieve and print a JSON-LD document."""
+    if is_stdin(input_):
```
mielvds, line 184: This is personal, but I prefer not having stdin as default input and always having to explicitly set input to stdin by adding `-`. To some, running `pyld get` makes it seem like the process is hanging. I'd expect the help, error message or both instead.

tests/test_cli.py
```diff
@@ -0,0 +1,273 @@
...
+def test_expand_from_stdin():
+    result = runner.invoke(cli.pyld, ['expand', '-'], input=json.dumps(PERSON))
+    assert result.exit_code == 0, result.output
+    assert json.loads(result.stdout) == [
+        {'http://schema.org/name': [{'@value': 'Ada Lovelace'}]},
+    ]
+
+
+def test_expand_with_base(tmp_path: Path):
+    doc = {
```
mielvds: This is minor, but we consistently used `input` as variable name in test_jsonld.py

tests/test_cli.py
```diff
@@ -0,0 +1,273 @@
...
+    )
+    assert result.exit_code == 0, result.output
+    assert json.loads(result.stdout) == [
+        {'http://schema.org/name': [{'@value': 'Ada Lovelace'}]},
+    ]
+
+
+def test_expand_with_context_file(tmp_path: Path):
+    context_path = tmp_path / 'context.jsonld'
+    context_path.write_text(
```
mielvds: consider a helper function for this path + write_text pattern

mielvds (approved):
Nice work @anatoly-scherbakov , some minor comments. FYI, I won't publish v3.3.0 before #306, since the CLI is pretty incomplete now.

mielvds (approved):
Looks great 👍

PR #330, opened 2026-08-28
Complete PyLD CLI transformations

anatoly-scherbakov:
## Summary

- Add `compact`, `flatten`, `frame`, `to-rdf`, and `from-rdf` commands.
- Split the CLI into an application module, shared helpers, and per-command modules.
- Add runnable command documentation and deterministic fixtures.

## CLI contract

- JSON-LD inputs accept a local path, HTTP(S) URL, or `-` for standard input.
- Context and frame operands accept a local path, HTTP(S) URL, or `-`; only one operand may read standard input.
- N-Quads input accepts a local path or `-`; N-Quads output is written without terminal styling.
- The CLI supports JSON-LD 1.1 only.

## Validation

- `pytest -q tests/test_cli.py` (76 passed)
- `make lint`


lib/pyld/options.py
```diff
@@ -121,8 +121,8 @@ class FrameOptions(ExpandContextOptions, total=False):
     expandContext: Context
     """A context to expand with."""
 
-    embed: Literal['@last', '@always', '@never', '@link']
-    """Default `@embed` flag (default: `@last`)."""
+    embed: Literal['@once', '@last', '@always', '@never', '@link']
```
mielvds, line 124: This was probably originally taken from: https://github.com/digitalbazaar/pyld/blob/263af94fda2300fce52a2449d93f0d823b57f609/lib/pyld/jsonld.py#L744. Could you fix the docstrings too?

tests/test_cli.py
```diff
@@ -171,6 +204,433 @@ def test_expand_requires_input():
     assert 'Missing argument' in result.output
 
 
+def test_cli_is_jsonld_11_only():
```
mielvds, line 218: These tests don't have docstrings; I missed that in the previous PR. Could you add them?

mielvds (changes requested):
Very nice! Just a request for docstrings :)

mielvds (approved):

PR #338, opened 2026-09-13
Copy @direction when cloning the active context; guard the @direction null reset

### LINKED ISSUE #337: Default @direction is not inherited across context layers — _clone_active_context omits @direction
## Summary

The active-context clone used by context processing copies `@base`, `@language`, `@vocab`, and `previousContext`, but not `@direction`. As a result the default base direction is silently dropped whenever a new context layer is processed on top of an existing active context — a second document-level `@context` layer, a property-scoped context, a type-scoped context, an embedded node `@context`, or a remote context. The default language is inherited correctly in all of these cases; only `@direction` is lost.

This is the same omission as digitalbazaar/jsonld.js#586 — PyLD's `_clone_active_context` mirrors jsonld.js's `_cloneActiveContext`, so the two implementations should probably fix it together.

Reproduced on PyLD 3.3.0 (Python 3.12); `_clone_active_context` on current `master` has the same omission.

## Reproduction

```python
import json
from pyld import jsonld

# 1. Control: a single context layer sets both defaults — both apply.
print(json.dumps(jsonld.expand({
    '@context': {'@language': 'en', '@direction': 'rtl'},
    'http://ex/p': 'v'
})))

# 2. A second document-level context layer: @language survives, @direction is lost.
print(json.dumps(jsonld.expand({
    '@context': [{'@language': 'en', '@direction': 'rtl'}, {'dummy': 'http://ex/d'}],
    'http://ex/p': 'v'
})))

# 3. A property-scoped context: @language is inherited into the scope, @direction is lost.
print(json.dumps(jsonld.expand({
    '@context': {
        '@language': 'en',
        '@direction': 'rtl',
        'thing': {'@id': 'http://ex/thing', '@context': {'other': 'http://ex/other'}}
    },
    'thing': {'http://ex/label': 'hello'}
})))

# 4. to_rdf with rdfDirection: the divergence reaches N-Quads.
print(jsonld.to_rdf({
    '@context': {
        '@language': 'en',
        '@direction': 'rtl',
        'thing': {'@id': 'http://ex/thing', '@context': {'other': 'http://ex/other'}}
    },
    '@id': 'http://ex/x',
    'http://ex/out': 'outside',
    'thing': {'http://ex/label': 'hello'}
}, {'format': 'application/n-quads', 'rdfDirection': 'i18n-datatype'}))
```

Actual output (PyLD 3.3.0):

```
[{"http://ex/p": [{"@language": "en", "@direction": "rtl", "@value": "v"}]}]
[{"http://ex/p": [{"@language": "en", "@value": "v"}]}]
[{"http://ex/thing": [{"http://ex/label": [{"@language": "en", "@value": "hello"}]}]}]
<http://ex/x> <http://ex/out> "outside"^^<https://www.w3.org/ns/i18n#en_rtl> .
<http://ex/x> <http://ex/thing> _:b0 .
_:b0 <http://ex/label> "hello"@en .
```

Expected (per spec): case 2 keeps `"@direction": "rtl"`; case 3's `hello` carries `"@direction": "rtl"`; case 4 emits `"hello"^^<https://www.w3.org/ns/i18n#en_rtl>` for the in-scope literal. Note case 4: within one document, the top-level string gets the i18n datatype while the string inside the scope silently degrades to a plain language-tagged literal.

(jsonld.js 9.0.0 produces byte-identical output for all four cases — the two implementations share the bug.)

## Root cause

`_clone_active_context` in `lib/pyld/jsonld.py`:

```python
def _clone_active_context(self, active_ctx):
    """
    Clones an active context, creating a child active context.

    :param active_ctx: the active context to clone.

    :return: a clone (child) of the active context.
    """
    child = {'mappings': dict(active_ctx['mappings'])}
    if '@base' in active_ctx:
        child['@base'] = active_ctx['@base']
    if 'previousContext' in active_ctx:
        child['previousContext'] = active_ctx['previousContext']
    if '@language' in active_ctx:
        child['@language'] = active_ctx['@language']
    if '@vocab' in active_ctx:
        child['@vocab'] = active_ctx['@vocab']
    return child
```

`@direction` is never copied. Context processing starts each layer with `rval = self._clone_active_context(rval)`, so any layer processed after the one that set `@direction` loses it (unless that layer restates it — explicitly *setting* `@direction` works, because the `'@direction' in ctx` handler assigns onto the clone; only inheritance is broken).

## Spec reference

JSON-LD 1.1 API, [Context Processing](https://www.w3.org/TR/json-ld11-api/#context-processing-algorithm): the active context "consists of: the active term definitions …, the current base IRI, an inverse context, an optional vocabulary mapping, an optional default language, an optional **default base direction**, and an optional previous context", and step 1 of the algorithm is "Initialize result to the result of **cloning active context**, with inverse context set to `null`." Step 5.8 then only modifies the default base direction "If context has a `@direction` entry" — an absent entry leaves the inherited value in place.

A two-line fix in `_clone_active_context` appears sufficient:

```python
    if '@direction' in active_ctx:
        child['@direction'] = active_ctx['@direction']
```

## Notes

- **Same bug filed against jsonld.js**: digitalbazaar/jsonld.js#586 (with the equivalent one-line fix in `_cloneActiveContext`). Fixing one without the other would make the two implementations diverge from each other.
- **Implementations already diverge on this**: Titanium JSON-LD (Java) copies both `defaultLanguage` and `defaultBaseDirection` in its `ActiveContext` copy constructor, so it inherits `@direction` per spec and produces different expansion/N-Quads output for the documents above.
- **The W3C test suite doesn't cover this** — there is no json-ld-api fixture combining a default `@direction` with a subsequent context layer or scoped context, which is presumably why it has gone unnoticed in both implementations.
- **Compatibility note**: fixing this changes canonical N-Quads (and therefore RDFC hashes / proof values) for any signed document that combines a default `@direction` with multiple context layers or scoped contexts under `rdfDirection: 'i18n-datatype'`. Related standards discussion on `@direction`/`rdfDirection` divergence in Data Integrity: w3c/vc-data-integrity#366.

skydudie:
Fixes #337.

### What

Two fixes in `_process_context`-related direction handling, mirroring jsonld.js semantics per JSON-LD 1.1 API [Context Processing](https://www.w3.org/TR/json-ld11-api/#context-processing-algorithm):

1. **`_clone_active_context` now copies `@direction`**, matching the existing handling of `@language`/`@base`/`@vocab`. Previously any context layer processed on top of an existing active context — a second document-level layer, a property-/type-scoped context, an embedded node `@context`, or a remote context — silently dropped the inherited default base direction (while the default language survived), changing expanded output and, under `rdfDirection`, the emitted N-Quads. (#337; same bug as digitalbazaar/jsonld.js#586; matching PR digitalbazaar/jsonld.js#587.)
2. **A `@direction: null` context entry is now a no-op when no default direction is set** (`rval.pop('@direction', None)` instead of an unguarded `del`). This mirrors JS `delete` semantics. Found while writing the tests: on released 3.3.0, `jsonld.expand({'@context': {'@direction': None}, 'http://ex/p': 'v'})` raises `KeyError: '@direction'`, and a legitimate scoped `@direction: null` reset is rejected as `invalid scoped context`.

### Tests

- Three `TestExpand` methods (per AGENTS.md, added to the existing class): default `@direction` survival across document-level layers, inheritance into a property-scoped context, and scoped override / `null` reset.
- Full local run (`pytest`, W3C suites included via the `specifications/` submodules): **1885 passed / 18 failed / 29 skipped / 1 xfailed** on this branch vs **1882 / 18 / 29 / 1** on unpatched `master` — the failure lists are byte-identical (`tests/test_cli.py` JSON-decode issues and a missing `aiohttp` module in this environment), i.e. +3 new passing tests, zero regressions.
- `make lint` clean.

### Compatibility note

This changes expanded output (and N-Quads under `rdfDirection`) for documents that combine a default `@direction` with multiple context layers or scoped contexts — flagging for the 4.0.0 changelog since RDFC hashes over such documents change. Related: w3c/vc-data-integrity#366.


CHANGELOG.md
```diff
@@ -23,6 +23,8 @@
 ### Fixed
 
 - Preserved RDF literal lexical forms when converting through RDFLib, including canonical double output, large numeric values, and compound literal handling.
+- Copy the default base direction (`@direction`) when cloning the active context, matching the existing handling of the default language. Previously any context layer processed on top of an existing active context — a second document-level layer, a property-/type-scoped context, an embedded node `@context`, or a remote context — silently dropped the inherited `@direction` (while `@language` survived), changing expanded output and, under `rdfDirection`, the emitted N-Quads. [Issue #337](https://github.com/digitalbazaar/pyld/issues/337).
```
mielvds, line 26: This is too long and not phrased as a fix. Could you rephrase?

mielvds (commented):
Thanks @skydudie. Could you rebase against `codex/255-split-iri-and-util-tests`?

CHANGELOG.md
skydudie: Rephrased in https://github.com/digitalbazaar/pyld/commit/5b65a426a3c93bd3189c88eba46b743f8967f7b4 !

skydudie:
> Thanks @skydudie. Could you rebase against `codex/255-split-iri-and-util-tests`?

Rebased onto codex/255-split-iri-and-util-tests and retargeted the PR to it. Moved the tests from tests/test_jsonld.py into tests/jsonld/test_expand.py

Lmk if there is anything else ya want changed😃

CHANGELOG.md
mielvds: perfect, thx!

mielvds:
yes, could you rebase to master? :) I had a small fix in

skydudie:
> yes, could you rebase to master? :) I had a small fix in

Rebased onto master and thanks for the schema.org fix!

mielvds:
Thx @skydudie looks good. @anatoly-scherbakov could you have a final look at this?

anatoly-scherbakov (commented):
The W3C suites `@direction: null` inputs first establish a default direction, so they do not exercise the no-default-direction case that previously raised `KeyError`.

This focused test would cover the `pop(..., None)` behavior directly:

```python
def test_default_direction_null_reset_is_noop_when_unset():
    input = {
        "@context": {"@direction": None},
        "http://example.com/p": "v",
    }

    expected = [
        {
            "http://example.com/p": [
                {"@value": "v"},
            ],
        },
    ]

    assert jsonld.expand(input) == expected
```

skydudie:
> The W3C suites `@direction: null` inputs first establish a default direction, so they do not exercise the no-default-direction case that previously raised `KeyError`.
> 
> This focused test would cover the `pop(..., None)` behavior directly:
> 
> ```python
> def test_default_direction_null_reset_is_noop_when_unset():
>     input = {
>         "@context": {"@direction": None},
>         "http://example.com/p": "v",
>     }
> 
>     expected = [
>         {
>             "http://example.com/p": [
>                 {"@value": "v"},
>             ],
>         },
>     ]
> 
>     assert jsonld.expand(input) == expected
> ```

Good catch thanks @anatoly-scherbakov!  Added your test in https://github.com/digitalbazaar/pyld/pull/338/commits/cf845157700bbf167968167672b878e193a88f0e

anatoly-scherbakov (approved):

Issue #255, opened 2026-05-04
Review test structure


anatoly-scherbakov:
We need to discuss testing structure. Function-based tests, fine, but I don't want to end up with hundreds of AI generated tests that we can no longer navigate. And the very least we should stick to one file per function

_Originally posted by @mielvds in https://github.com/digitalbazaar/pyld/pull/249#pullrequestreview-4218798983_

anatoly-scherbakov:
I agree. I like smaller files more than bigger ones. @mielvds we could record these requirements in `AGENTS.md`.

mielvds:
Ideally, I'd like the tests file structure to mimic the code file structure, as is common. This is currently the case, but we probably need to split up `jsonld.py` more first before we can get something maintainable. 

How would you partition these function-based tests, @anatoly-scherbakov ? 
- One file per "problem" (e.g. regression tests)
- one file per python function
- one file per `lib/*/*.py` file?

mielvds:
@anatoly-scherbakov might want to do at least the move to function-based tests before 4.0.0 release.
How about the following directory structure:
```
tests/
  test_cli.py
  test_manifests.py

  document_loader
    test_document_loader.py
    test_file_document_loader.py
    test_frozen_document_loader.py
    test_scheme_directed_document_loader.py
    test_sqlite_cache_requests_document_loader.py
    test_type_directed_document_loader.py
  
  jsonld/
    test_expand.py
    test_frame.py
    test_flatten.py
    test_normalize.py
    test_to_rdf.py
    test_from_rdf.py
    test_compact.py

  iri_resolver/
    test_resolve.py
    test_unresolve.py
    test_remove_dot_segments.py

  util/
    test_to_legacy_dataset.py
    test_from_legacy_dataset.py
```

PR #340, opened 2026-09-13
[#255] Split JSON-LD tests by API

anatoly-scherbakov:
Part of #255; stacked on #339.

Replaces the combined JSON-LD test module with function-based modules grouped by public API.

Validation:
- `pytest tests/jsonld`
- `pytest --ignore tests/test_manifests.py`
- `make lint`


mielvds (commented):
Thanks for the action @anatoly-scherbakov ! May I suggest a slight scope change to this PR. In to_rdf, We have two nquads formatting styles mixed:
``` 
assert nquads == (
        '<http://example.com/s> <http://example.com/p> '
        '"1.0E21"^^<http://www.w3.org/2001/XMLSchema#double>  .\n\n'
    )
```

and 

```
expected = """_:custom0 <http://example.org/p> _:custom1  .
_:custom1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#first> "a"^^<http://www.w3.org/2001/XMLSchema#string>  .
_:custom1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#rest> _:custom2  .
_:custom2 <http://www.w3.org/1999/02/22-rdf-syntax-ns#first> "b"^^<http://www.w3.org/2001/XMLSchema#string>  .
_:custom2 <http://www.w3.org/1999/02/22-rdf-syntax-ns#rest> <http://www.w3.org/1999/02/22-rdf-syntax-ns#nil>  .

"""
```

Could we unify those to either one? This PR seems like the place to do this

anatoly-scherbakov:
Updated the complete exact N-Quads expectations in `test_to_rdf.py` to the named triple-quoted `expected` style, while retaining the sorted multiline comparisons, substring assertions, and existing TODOs. Focused tests and lint pass.

PR #345, opened 2026-09-21
Migrate packaging and development workflows to uv

### LINKED ISSUE #344: Migrate packaging and development workflows to `uv` and `pyproject.toml`
## Summary

Migrate PyLD from `setup.py` and separate requirements files to a modern `pyproject.toml`-based project managed with `uv`

gorewilliams:
## Summary

- move package metadata, dependencies, extras, entry points, and package data from `setup.py` into `pyproject.toml`
- replace requirements files with uv dependency groups and a committed `uv.lock`
- run local development, testing, linting, documentation, and release workflows through uv
- update GitHub Actions and Dependabot for locked uv environments
- update contributor and installation documentation

## Validation

- `uv lock --check`
- `make lint`
- `make test` — 1,903 passed, 27 skipped, 1 xfailed
- `uv build --no-sources`, with wheel and sdist contents verified
- `make docs-build` in strict mode
- local versioned-documentation deployment rehearsal

Closes #344


README.md
```diff
@@ -357,11 +357,11 @@ at:
 
 [https://github.com/digitalbazaar/pyld](https://github.com/digitalbazaar/pyld)
 
-You can install the source using `make`:
+Install [uv](https://docs.astral.sh/uv/), then create the locked development
```
mielvds, line 360: `make install` still exists so why not keep it?

README.md
```diff
@@ -409,7 +409,7 @@ PyLD source directory, then all unittests, including the sample test suites, can
 be run with `pytest`:
 
 ```bash
-pytest 
+uv run --locked pytest
```
mielvds, line 412: `uv run pytest` should be enough?

README.md
gorewilliams: I wanted to emphasize the need to have `uv` installed. And honestly, since `make install` is now a one-liner `uv sync --locked` - it just seemed clearer to me. Can change this if you disagree.

README.md
gorewilliams: The idea with `--locked` is that we want to guarantee that the correct set of dependencies is installed at runtime, and that there isn't a conflict between `pyproject.toml` and `uv.lock`.

anatoly-scherbakov:
I agree with feedback by @mielvds, with that said, looks good to me, thanks!

gorewilliams:
Thx! Wdyt @mielvds?

.github/workflows/main.yaml
```diff
@@ -91,19 +92,19 @@ jobs:
...
+          version: '0.12.17'
+          enable-cache: true
       - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install -r requirements.txt
-      - name: Install testing dependencies
-        run: |
-          pip install -r requirements-test.txt
+        run: uv sync --locked
```
mielvds, line 101: @gorewilliams are test dependencies always installed now?

mielvds:
I rebased and added a comment on make install. I see that rdflib also moved away from requirements.txt, so I think it's ok to no longer keep those. Thanks for the contribution, will merge soon.

.github/workflows/main.yaml
gorewilliams: Yes. Test deps are included in the default `dev` group.

gorewilliams: They're always installed when using `uv sync` plainly, as `dev` is the default group.

mielvds: Understood!

